PR #27, opened 2016-10-19
Use the NDK at API level 16 for backward compatibility with Jellybean

jason-simmons:


jason-simmons:
Revised this because x86-64 targets need a different NDK

PR #41, opened 2016-11-30
Update license script to handle recent changes.

Hixie:


Hixie:
cc @abarth

abarth:
LGTM

abarth:
We should move the license script into engine.git at some point.

PR #43, opened 2016-12-02
Install android tools on Travis.

Hixie:
We need these to determine the licenses that apply.


Hixie:
cc @jason-simmons

jason-simmons:
LGTM

PR #45, opened 2017-01-21
Make some hooks portable to run on Windows.

goderbauer:
* tools/dart/update.py can now also update the Dart SDK on Windows
* tools/android/download_android_tools.py does not error on Windows anymore (it exits gracefully with a notice)

The Android Tools are not needed on a Windows host because we are prebuilding the Android binaries on a Linux host.

required for flutter/flutter#138


jason-simmons:
LGTM

PR #46, opened 2017-01-23
Add portable buildtools update script

goderbauer:
It will replace buildtools/update.sh in the DEPS hooks of the engine.


abarth:
LGTM

jason-simmons:
LGTM

PR #53, opened 2017-03-14
Strip symbols of clang_x86 binaries

jason-simmons:
The clang_x86 toolchain is used by gen_snapshot in profile/release build modes


jason-simmons:
@abarth @chinmaygarde

abarth:
LGTM

jason-simmons:
Yes - the gen_snapshot binaries currently distributed to bin/cache/artifacts are unstripped

PR #71, opened 2017-06-07
Set ios_deployment_version to iOS 8

cbracken:


cbracken:
Related to flutter/flutter#10510

PR #76, opened 2017-06-28
Don't set Unix file permissions when extracting Dart SDK on Windows

goderbauer:
Prevents from setting dart.exe incorrectly to read-only which causes failures when the bot tries to delete it.


goderbauer:
Should fix: https://build.chromium.org/p/client.flutter/builders/Windows%20Engine/builds/420/steps/gclient%20runhooks/logs/stdio

PR #77, opened 2017-06-29
Checkout prebuilt dart into standalone dart location.

aam:
If we check out prebuilt dart executables into the same location where standalone dart sdk expects it, we don't need to customize that location for flutter vs dart. This simplifies scripts and fixes the problem where for andorid flutter build patched_sdk step doesn't use prebuilt dart binary, uses [slow] dart binary that targets arm, runs via simulator.

### DIFF
--- a/tools/dart/update.py
+++ b/tools/dart/update.py
@@ -27,9 +27,27 @@ WINDOWS_64_SDK = 'dartsdk-windows-x64-release.zip'
 
 # Path constants. (All of these should be absolute paths.)
 THIS_DIR = os.path.abspath(os.path.dirname(__file__))
-MOJO_DIR = os.path.abspath(os.path.join(THIS_DIR, '..', '..'))
-DART_SDKS_DIR = os.path.join(MOJO_DIR, 'dart/tools/sdks')
-PATCH_FILE = os.path.join(MOJO_DIR, 'tools', 'dart', 'patch_sdk.diff')
+FLUTTER_DIR = os.path.abspath(os.path.join(THIS_DIR, '..', '..'))
+DART_SDKS_DIR = os.path.join(FLUTTER_DIR, 'dart/tools/sdks')
+PATCH_FILE = os.path.join(FLUTTER_DIR, 'tools', 'dart', 'patch_sdk.diff')
+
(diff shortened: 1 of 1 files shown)


tools/dart/update.py
```diff
@@ -28,54 +28,55 @@
 # Path constants. (All of these should be absolute paths.)
 THIS_DIR = os.path.abspath(os.path.dirname(__file__))
 MOJO_DIR = os.path.abspath(os.path.join(THIS_DIR, '..', '..'))
```
zanderso: Maybe rename MOJO_DIR -> FLUTTER_DIR while you're in here.

tools/dart/update.py
```diff
@@ -73,8 +74,11 @@ def main():
     with zipfile.ZipFile(output_file, 'r') as zip_ref:
```
zanderso: Small cleanup suggestion: Maybe pull some or all of the zipfile extraction into a helper function.

tools/dart/update.py
```diff
@@ -28,54 +28,55 @@
...
         get_sdk = True
 
   if get_sdk:
     # Completely remove all traces of the previous SDK.
-    if os.path.exists(DART_SDK_DIR):
-      shutil.rmtree(DART_SDK_DIR)
-    os.mkdir(DART_SDK_DIR)
+    if os.path.exists(dart_sdk_dir):
+      shutil.rmtree(dart_sdk_dir)
+    os.mkdir(dart_sdk_dir)
```
zanderso: I think this removal logic is no longer pointing at the right path. It should be removing dart_sdk_dir + 'dart-sdk'.

tools/dart/update.py
aam: Done.

tools/dart/update.py
aam: Done.

tools/dart/update.py
aam: Done.

aam:
Yes, the plan was to get https://codereview.chromium.org/2963053002/ at least lgtm'ed too before pinging flutter folks. Without that cl this won't work.

aam:
Thanks!

PR #88, opened 2017-08-04
Set is_flutter flag to true

aam:
This sets the flag proposed in https://codereview.chromium.org/2998503002.


zanderso:
Same comment here as on the Dart CL.

PR #92, opened 2017-08-10
Fix windows build rules so that we can build dart sdk as part of Flutter engine.

aam:
Few problems were revealed as a result of failed attempt to land https://chromium.googlesource.com/external/github.com/flutter/engine/+/a5bac1ee09668f4687dc81c74bcf8adcad85cd7c here
https://build.chromium.org/p/client.flutter/builders/Windows%20Engine/builds/538.

The solution implemented in this PR is to take the code from corresponding Dart build/toolchain.

@bkonyi


aam:
Thanks!

PR #113, opened 2017-11-01
Always use the x86_64 Android toolchain.

zanderso:
This is a step towards removing the dart_host_toolchain GN argument. This selects the Android toolchain that runs on x86_64 even if host_cpu is x86. This matches what Dart does in its GN config for Android.

### DIFF
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -54,10 +54,8 @@ if (is_android) {
 
   # Defines the name the Android build gives to the current host CPU
   # architecture, which is different than the names GN uses.
-  if (host_cpu == "x64") {
+  if (host_cpu == "x64" || host_cpu == "x86") {
     android_host_arch = "x86_64"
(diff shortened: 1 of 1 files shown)


build/config/android/config.gni
```diff
@@ -54,10 +54,8 @@ if (is_android) {
 
   # Defines the name the Android build gives to the current host CPU
   # architecture, which is different than the names GN uses.
-  if (host_cpu == "x64") {
+  if (host_cpu == "x64" || host_cpu == "x86") {
```
aam: This means that build should always be done on 64-bit host, right, while the removed code theoretically allowed you to build on 32-bit  host?

zanderso: In theory, yes. In practice the 32-bit toolchain doesn't actually work, and we don't have any 32-bit-only hosts.

aam: okay, sgtm

PR #115, opened 2017-11-08
Temporarily disable -Wtautological-constant-compare

cbracken:
Temporarily disable this warning because the current implementation is
too aggressive and is reporting cases that are generally safe.

Revert this once https://reviews.llvm.org/D39462 lands.


cbracken:
Related change in Fuchsia: https://fuchsia.googlesource.com/build/+/master/config/BUILD.gn#121

PR #121, opened 2017-12-31
We really don't require openbox to build Flutter.

Hixie:


Hixie:
This changed my active window manager. I uninstalled openbox and build everything successfully, so I don't think we need it.

yjbanov:
lgtm

PR #123, opened 2018-01-18
Roll dart-sdk to 2.0.0-dev.16.0

aam:
This is follow-up to https://github.com/flutter/engine/pull/4554 to fix https://build.chromium.org/p/client.flutter buildbots.

### DIFF
--- a/tools/dart/update.py
+++ b/tools/dart/update.py
@@ -19,7 +19,7 @@ import zipfile
 # How to roll the dart sdk: Just change this url! We write this to the stamp
 # file after we download, and then check the stamp file for differences.
 SDK_URL_BASE = ('http://gsdview.appspot.com/dart-archive/channels/dev/raw/'
-                '1.25.0-dev.11.0/sdk/')
+                '2.0.0-dev.16.0/sdk/')
 
 LINUX_64_SDK = 'dartsdk-linux-x64-release.zip'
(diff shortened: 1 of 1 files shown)


Hixie:
Huh. Why aren't we just using the Dart version from the Dart SDK (i.e. the hash files that are in the target directory)? That's weird.

Hixie:
In any case, LGTM.

aam:
@Hixie wrote 
> Huh. Why aren't we just using the Dart version from the Dart SDK (i.e. the hash files that are in the target directory)? That's weird.

You mean why we are not downloading dart sdk per `dart_revision` from DEPS?
I guess it should be possible.
Alternatively, we could use dart sdk that we already build when we target "host": https://github.com/flutter/engine/blob/master/BUILD.gn#L19

Hixie:
What I mean is that the directory into which we put this SDK has a hash in it, so shouldn't we just use the version described by that hash?

Hixie:
e.g. https://github.com/dart-lang/sdk/blob/master/tools/sdks/linux/dart-sdk.tar.gz.sha1

aam:
Ah, I didn't realize there is sha1 in those folders. It might be having version hardcoded explicitly in `update.py` goes back to the time when flutter wanted to stay on older version of dart sdk analyzer, so it's not affected by dart rolls.

PR #126, opened 2018-02-13
Enable -Wunguaraded-availability for iOS builds

cbracken:
Ensures we don't accidentally add any dependencies on iOS APIs > iOS 8
without putting them behind an @available guard.

Note that -Wunguarded-availability implies the older
-Wpartial-availability (which is actually an alias for it in recent
clang toolchains).


cbracken:
Related flutter/engine#4664

cbracken:
Done!

PR #127, opened 2018-02-13
Eliminate is_nacl from build config

cbracken:
Given our lack of use of NaCL and its deprecation, eliminating it seems
like generally not a bad idea.


cbracken:
Of course... please shout if you can think of a reason this *is* a bad idea :)

PR #129, opened 2018-03-12
Add an ignore rule for googletest

jason-simmons:


cbracken:
Would moving libtxt out of flutter/third_party help? Then we could just do a blanket third_party/*

jason-simmons:
There are still some third_party/ files that live within the buildroot repository:
    https://github.com/flutter/buildroot/tree/master/third_party

PR #130, opened 2018-03-12
Obtain the toolchain version based on the head commit of buildtools

jason-simmons:
This will be required for future Fuchsia toolchains that are based on the
CIPD tool and do not write a clang.stamp file


jason-simmons:
@cbracken @chinmaygarde

PR #131, opened 2018-04-03
Derive iOS host toolchain from $host_cpu

cbracken:
Required to build 32-bit gen_snapshot for macOS.


cbracken:
Related: flutter/flutter#2089

PR #138, opened 2018-05-15
Update Windows version of GN based on that of Dart.

rmacnak-google:


tools/buildtools/update.py
```diff
@@ -25,7 +26,8 @@ def Update():
 
 
 def UpdateOnWindows():
-    sha1_file = os.path.join(BUILDTOOLS, 'win', 'gn.exe.sha1')
```
rmacnak-google: The old path here is deleted in upstream Fuchsia buildtools.

PR #145, opened 2018-05-30
update bucket settings

jonahwilliams:
Moving android tools to our own bucket.  

I believe this update is safe, since the version has to be incremented separately in flutter/engine DEPS file


jonahwilliams:
I also have https://github.com/flutter/flutter/pull/18077 pending now.

jonahwilliams:
Not working on this now

PR #159, opened 2018-08-31
Update Android NDK to r17b

goderbauer:


goderbauer:
I verified that Flutter Gallery still runs on Phones with API level 16 after this change.

PR #165, opened 2018-09-10
Enable LTO for Android

goderbauer:
Size reductions of libflutter.so:
* uncompressed: -412KB (from 6MB to 5.6MB)
* compressed: -243KB (from 3.3MB to 3MB)

Fixes https://github.com/flutter/flutter/issues/16538.

### DIFF
--- a/build/config/compiler/BUILD.gn
+++ b/build/config/compiler/BUILD.gn
@@ -17,6 +17,7 @@ if (is_win) {
 }
 
 import("//build/toolchain/ccache.gni")
+import("//build/toolchain/clang.gni")
 import("//build/config/profiler.gni")
 import("//build/config/sanitizers/sanitizers.gni")
 
@@ -797,7 +798,7 @@ if (is_win) {
     common_optimize_on_ldflags += [
       # Specifically tell the linker to perform optimizations.
       # See http://lwn.net/Articles/192624/ .
-      "-Wl,-O1",
+      "-Wl,-O2",
       "-Wl,--gc-sections",
     ]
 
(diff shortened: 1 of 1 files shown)


build/config/compiler/BUILD.gn
```diff
@@ -812,16 +813,23 @@ if (is_win) {
...
 # Default "optimization on" config. On Windows, this favors size over speed.
 config("optimize") {
+  lto_flags = []
   if (is_win) {
     # Favor size over speed, /O1 must be before the common flags. The GYP
     # build also specifies /Os and /GF but these are implied by /O1.
     cflags = [ "/O1" ] + common_optimize_on_cflags + [ "/Oi" ]
-  } else if (is_android || is_ios) {
+  } else if (is_ios) {
     cflags = [ "-Os" ] + common_optimize_on_cflags  # Favor size over speed.
```
goderbauer: Using `-Oz` on iOS builds actually increases size, even though clang doc says: 

```
-Oz   Like -Os (and thus -O2), but reduces code size further.
```

cbracken: Interesting - @petrhosek might have more info.

cbracken:
Please keep an eye on the benchmarks after this rolls into the framework.

cbracken:
lgtm

goderbauer:
> Please keep an eye on the benchmarks after this rolls into the framework.

Will do!

build/config/compiler/BUILD.gn
petrhosek: This is not entirely surprising. LTO usually leads to size increase due to inlining and it currently doesn't currently take `-Oz` into account. We had some discussions about this in the past, but the conclusion is that it's not quite clear what LTO with `-Oz` should do since those are to some extent conflicting options.

goderbauer: Thanks for the explanation!

PR #186, opened 2018-10-29
Revert "Add compiler flags suggested by security review (#4368) (#177)"

goderbauer:
This reverts commit 9bbd43537961a12b1a4f164cb0dc2c52e5ee6d31.

This caused major regressions, see https://github.com/flutter/flutter/issues/23678.

In the future, we can check the flags individually to see which we can add without regressing our benchmarks.


kf6gpe:
(late, but LGTM.) Thanks!!

PR #194, opened 2018-11-21
Roll Android SDK to 28

dnfield:
- Adds instructions for rolling the Android SDK/NDK
- Adds script for uploading the Android SDK/NDK to cloud storage
  - One problem with this: I couldn't get it to work with depot_tools, it wouldn't use my credentials.  I kept having to manually use gsutil to actually upload.
- Rolls Android SDK to 28
- Rolls Android Build Tools to 28.0.3
- Rolls Android support library to v13
- Uses d8 for dexing instead of dx (required for Java8 support)
- Removes the play services resource references.  AFAICT these are all legacy bits from Chromium and we don't use/need them.  They're also harder to pull in with newer versions of the SDK because they're now contained in AAR files .. but I can't find anywhere that they're used or exposed anyway.

Will require the following changes in flutter/engine: 
- Replace `v4` with `v13` in `shell/platform/android/BUILD.gn`

~~And we'll need to update the `build.gradle` template(s) in flutter/flutter to include this:~~ (Done)

```
android {
    compileSdkVersion 27 // 28?

    compileOptions {
        sourceCompatibility 1.8
        targetCompatibility 1.8
    }
```

/cc @mklim @matthew-carroll @chinmaygarde @jason-simmons @cbracken @jonahwilliams @goderbauer


cbracken:
Thanks for working on this!

I haven't taken a look yet, but unless there's a reason to land this urgently, we should hold off rolling this into the engine for a couple weeks until we've cut our next release.

dnfield:
Requisite changes in `flutter/flutter` have landed.

@cbracken - can you take a look at this at this point?

dnfield:
Fixes https://github.com/flutter/flutter/issues/20717

PR #220, opened 2019-02-26
Add clang static analysis support

cbracken:
Adds support for building with clang static analysis enabled.


cbracken:
Tested on macOS. I still need to verify on Linux. There'll be a followup PR to add the option to `gn` in the engine.

cbracken:
Alright verified on Linux as well.

PR #223, opened 2019-03-21
Add Ubuntu 18 (bionic) support to install_build-deps scripts

stuartmorgan:


stuartmorgan:
These are minimal tweaks to get the scripts running on 18.04. I do see an odd link error on android_release_unopt, so 18.04 isn't necessary completely working, but this is at least most of the way there, and does allow the scripts to run (vs erroring out immediately).

PR #226, opened 2019-03-27
Add dependent libraries to GLFW

stuartmorgan:
The initial GLFW build rules did not include adding the necessary
dependent libraries to the link step.


stuartmorgan:
Also includes a `gn format` I forgot to run when initially landing it.

PR #230, opened 2019-04-03
Add libgtk-3-dev to dev_list

hugovdm:
While I'm not sure if this would be the right fix (i.e. for everyone), but without it my build was broken when I simply followed instructions, and this fixes it.


stuartmorgan:
This is my fault; I didn't update the docs when the Linux shell landed. I've update the setup page to include a new step 12:
"If you're on Linux, run: sudo ./flutter/build/install-build-deps-linux-desktop.sh"

There's been some discussion of re-aligning install-build-deps.sh with the current Chromium version, which would eliminate the need for that script, but for now the new script is part of the required setup.

cbracken:
Given the instructions have been updated and these scripts are primarily intended for our CI bots, and not as general-purpose scripts for preparing machines for engine builds, shall we close this PR?

stuartmorgan:
Yes, we can close this since it's redundant with the new script.

PR #232, opened 2019-04-09
Add macOS build rules for third_party/glfw

stuartmorgan:


stuartmorgan:
This is to allow bringing up an optional build of the GLFW shell on macOS (primarily for ease of development/testing of the GLFW shell)

PR #234, opened 2019-04-16
Updated Dart autoroller scripts

bkonyi:
Updated Dart autoroller scripts to:

- Respect TERM=dumb (don't print colors)
- Kill child processes properly if the roller is stopped
- Unbuffer output for better logging
- Allow for skipping the creation of a pull request
- Handle failing engine builders while waiting for pull request checks to pass
- Other mysterious features and bug fixes!

### DIFF
--- a/tools/dart/dart_autoroller.py
+++ b/tools/dart/dart_autoroller.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python -u
 # Copyright 2019 The Dart project authors. All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
@@ -41,8 +41,10 @@ from dart_roll_utils import *
 from git import Repo
 from github import Github, GithubException
 import argparse
+import atexit
 import datetime
 import os
+import signal
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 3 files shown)


tools/dart/dart_autoroller.py
```diff
@@ -272,7 +293,9 @@ def wait_for_status(commit):
     status = commit.get_combined_status().state
     if status == GITHUB_STATUS_SUCCESS:
       break
-    elif status == GITHUB_STATUS_FAILURE:
+    # If the only the engine builders are red, keep trying.
```
a-siva: if only the engine ....

tools/dart/dart_autoroller.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python -u
```
cbracken: Why hardcode the path? Not that we build on FreeBSD, and yes this is super specific to our infra, but this would break people whose python is at e.g. `/usr/local/bin/python`

https://stackoverflow.com/questions/2429511/why-do-people-write-the-usr-bin-env-python-shebang-on-the-first-line-of-a-pyt

bkonyi: I _think_ the existing string was breaking things with the '-u' option, but it seems fine now... I'll change it back.

PR #235, opened 2019-04-18
Build host Dart binary with -march=core2 on Mac OS X.

mraleph:
This is the right way to disable SSE3 features and above to support
early pre-Penryn iMac-s.

-msse2 which I tried to use before does not actually disable SSE3 and above.

This time I have verified that no pextrb (SSE4.1) instructions are generated
by C++ compiler by disassembling dart binary.

Note: the binary still contains some SSE4.1 instructions because BoringSSL comes
with handwritten assembly - but it check CPU features before using that assembly.

Fixes https://github.com/flutter/flutter/issues/24916

(This time for real)

### DIFF
--- a/build/config/compiler/BUILD.gn
+++ b/build/config/compiler/BUILD.gn
@@ -177,7 +177,7 @@ config("compiler") {
       common_mac_flags += [
         "-arch",
         "x86_64",
-        "-msse2",
+        "-march=core2",
       ]
     } else if (current_cpu == "x86") {
       common_mac_flags += [


cpalanzo:
Can someone tell me when this will be available in the SDK for installation? I see a lot of automated processes but could not find this available yet.

mraleph:
@cpalanzo I have not merged it yet :) I am gonna leave a comment here once it is all the way through the pipeline.

cpalanzo:
@mraleph Thank you! I appreciate the guidance - and am very impressed by the response to this issue and the quickness with which it get's resolved, reviewed, patched and pushed out - very very nice!

cpalanzo:
@mraleph So it looks like this was committed to the Flutter Engine - but is still not in the SDK? (I'm trying to figure out the process and appreciate you taking time to educate me.)

mraleph:
@cpalanzo I think now it should have reached master. Could you try again?

cpalanzo:
Yes - this works now! Thank you again for the assistance - I'm very impressed by the support within the flutter community and appreciate your direct help.

naclcaleb:
Happy to see this fixed! It was too late to be of use to me, but great work!

PR #236, opened 2019-04-24
Added --delete_removed_deps flag to create_updated_flutter_deps.py

bkonyi:
Added --delete_removed_deps flag to create_updated_flutter_deps.py to allow for dependencies removed in the Dart SDK to be automatically removed from flutter/DEPS.

### DIFF
--- a/tools/dart/create_updated_flutter_deps.py
+++ b/tools/dart/create_updated_flutter_deps.py
@@ -60,6 +60,10 @@ def ParseArgs(args):
       type=str,
       help='Flutter DEPS file.',
       default=os.path.join(FLUTTER_ROOT, 'DEPS'))
+  parser.add_argument('--delete_removed_deps',
+                      help='Automatically remove deleted dependencies.',
+                      action='store_true',
+                      default=False)
   return parser.parse_args(args)
 
 def Main(argv):
@@ -86,8 +90,8 @@ def Main(argv):
   lines = file.readlines()
   i = 0
   while i < len(lines):
-    updatedfile.write(lines[i])
     if lines[i].startswith("  'dart_revision':"):
+      updatedfile.write(lines[i])
       i = i + 2
       updatedfile.writelines([
         '\n',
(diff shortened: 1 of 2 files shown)


tools/dart/create_updated_flutter_deps.py
```diff
@@ -96,11 +100,31 @@ def Main(argv):
...
+        var_line = lines[i + 1]
+        for missing in missing_deps:
+          if missing in var_line:
+            write_dep = False
+            break
+      if write_dep:
+        updatedfile.writelines(lines[i:i+2])
+        updatedfile.write('\n')
+
+        # Handle case where there's a newline after each dependency key-value
```
aam: Perhaps you need to add similar 'WARNING: DO NOT EDIT MANUALLY' comment for this section too. Then you don't have to worry about newlines not being there.

bkonyi: It's probably easier just to check for the newline. This section is quite large and a message like that could be easily missed if someone does a find-and-goto like search.

aam: I think having 'WARNING: DO NOT EDIT MANUALLY' is important as it makes it clear this section is automatically generated/updated, consistently with section above it.

bkonyi: The issue is that this section does need to be edited manually when there's a new dependency being added. This section isn't updated with new entries by this script, but only removes entries when their variable has been deleted.

aam: I think more robust solution would be to extract, filter out removed dart deps, and print them out explicitly between "do not edit" tags same way we print out dart vars.

bkonyi: I don't know if that's really worth doing, especially given the format of this section. There's no consistent "marker" that prefaces the section of Dart deps that we can check for to print the message like we do above with the `dart_revision` which means we'd need extra look-ahead logic just to check if the message is already there so we don't print it multiple times to the file. Also, this section _does_ need to be edited manually when new dependencies are added so adding a message along the lines of "DO NOT EDIT" is incorrect anyway.

We should really avoid making assumptions about the whitespace and general formatting of the `deps` map as this section is manually edited from time to time, so I believe that this is the best approach to ensure formatting is consistent.

aam: I think it is worth extracting dart deps into separate section and come up with some kind of marker similar to how it was done for the section above.

bkonyi: I still don't think that this message will have any real value given that the section does need to be edited manually and will only add unnecessary complexity and brittleness if/when someone makes a bad manual change.

It makes sense for the `vars` section as edits made manually to tags / revisions in flutter/DEPS will be overwritten by the script, but that's not the case for the `deps` section as this change doesn't directly refer to the `vars` section of the Dart DEPS file to make removal decisions. If/when the script is updated to also add to the `deps` section based on new entries in the Dart DEPS file, then this might be worth revisiting.

aam: We are always rewriting both dart sections(`deps` and `vars`) when running this script - we are  using python parsing to parse python lines, we rely on that for `vars`, we should do the same for `deps`. Then we use well-defined format to write them out. If somebody used different formatting when adding new deps, that should not affect the script. The only thing we need to fix is delimiters between which this script has to insert dart deps. That seems pretty straightforward and robust.

cbracken:
@bkonyi do you plan to proceed with this PR?

bkonyi:
> @bkonyi do you plan to proceed with this PR?

Nope, I'll abandon it.

PR #250, opened 2019-05-08
Remove unused script

dnfield:
As of https://github.com/flutter/engine/commit/3671182a9533945ac8957c6bc68f164ca02df5b0, we no longer use this script - we directly get Dart via CIPD as a dep.


dnfield:
@bkonyi - can you confirm that this doesn't do anything bad to the autoroller?

bkonyi:
LGTM. This isn't used by the autoroller, but thanks for checking!

PR #256, opened 2019-05-17
Add some missing Fontconfig headers

jason-simmons:


jason-simmons:
While setting up Fontconfig I had accidentally placed these headers in the `third_party/fontconfig/src` directory (which is taken from a repository managed by Chromium).  So Fontconfig was building locally on my setup but failed when others tried to built it.

These headers need to be put in `third_party/fontconfig/include/src` (which is taken from Flutter's buildroot).

cbracken:

PR #259, opened 2019-06-05
Dart SDK Autoroller changes and bug fixes

bkonyi:
- Fixed syntax error
- Fixed global variables for repository paths not being overridden
correctly
- If the roll terminates while a PR is outstanding, the PR is now closed
- Pass --full-dart-sdk to gn.py to ensure that the web tooling is built
as well
- Update commit hashes to include dart-lang/sdk@ prefix in the roll
commit to link directly to GitHub commits.
- Removed --coverage for flutter test runs as it appears to sometimes
cause certain smoke tests to never complete.


bkonyi:
This change should prevent issues similar to those found in flutter/engine#9194 from reoccurring.

PR #279, opened 2019-07-18
Fuchsia CIPD entries will be tagged with git hash

iskakaushik:


iskakaushik:
superceded by https://github.com/flutter/buildroot/pull/280

PR #283, opened 2019-07-29
[bugfix] Fix fuchsia target arch selector

iskakaushik:


iskakaushik:
@chinmaygarde merging this as its trivial.

PR #286, opened 2019-07-30
Do not use neon instruction set of arm64 for fuchsia

iskakaushik:


iskakaushik:
@chinmaygarde LGTM'd in person.

PR #290, opened 2019-08-06
Add support for ANGLE using upstream build rules

stuartmorgan:
Reland of fe3b82877ee9dc7de7c345b8e1da1b0dd75f1a0a (#276)

This adds minimial (in some cases dummy) files sufficient to allow
building ANGLE in this buildroot using the build_with_chromium option.


stuartmorgan:
I have a license script fix ready for the issue that caused the revert; I'll follow this up right after landing with a roll of the buildroot in the engine so that it won't have the potential to break unrelated rolls like last time.

PR #291, opened 2019-08-06
Roll to Android SDK 29

shihaohong:
Updates the default android SDK version to 29 to accommodate changes to gestural navigation in Android Q. See https://github.com/flutter/engine/pull/10413/ for desired change.

### DIFF
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -17,7 +17,7 @@ if (is_android) {
   if (!defined(default_android_sdk_root)) {
     default_android_sdk_root = "//third_party/android_tools/sdk"
     default_android_sdk_version = "29"
-    default_android_sdk_build_tools_version = "28.0.3"
+    default_android_sdk_build_tools_version = "29.0.1"
   }
 
   declare_args() {
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -16,7 +16,7 @@ if (is_android) {
 
   if (!defined(default_android_sdk_root)) {
     default_android_sdk_root = "//third_party/android_tools/sdk"
-    default_android_sdk_version = "28"
+    default_android_sdk_version = "29"
     default_android_sdk_build_tools_version = "28.0.3"
   }
 


build/config/android/config.gni
```diff
@@ -16,7 +16,7 @@ if (is_android) {
 
   if (!defined(default_android_sdk_root)) {
     default_android_sdk_root = "//third_party/android_tools/sdk"
-    default_android_sdk_version = "28"
+    default_android_sdk_version = "29"
     default_android_sdk_build_tools_version = "28.0.3"
```
dnfield: Shouldn't we update the build tools as well?

dnfield:
I asked this question on the engine PR, but for posterity - how confident are we that using 29 is safe right now?

shihaohong:
@dnfield I'm unsure how safe it is and what checks we need to have, but in an ideal world, we'd like to be able to support gestural navigation early enough that app developers can adapt their Flutter apps. When the Android SDK was previously updated, what did we do to ensure that the update was stable enough?

build/config/android/config.gni
shihaohong: I'm not sure which version of the build tools go with this Android SDK. Is there some resource we can reference for this? I pulled the latest version of the build tools from Android Studio after installing SDK version 29

dnfield: Looks like 29.0.2 is available now.

build/config/android/config.gni
```diff
@@ -17,7 +17,7 @@ if (is_android) {
   if (!defined(default_android_sdk_root)) {
     default_android_sdk_root = "//third_party/android_tools/sdk"
     default_android_sdk_version = "29"
-    default_android_sdk_build_tools_version = "28.0.3"
+    default_android_sdk_build_tools_version = "29.0.1"
```
dnfield: Not 29.0.2?

shihaohong: I think the latest version of the SDK build tools is 29.0.1 - https://androidsdkmanager.azurewebsites.net/Buildtools

shihaohong: The [platform-tools](https://developer.android.com/studio/releases/platform-tools) are at v29.0.2

dnfield: Doh! You're right.  Thanks!

stuartmorgan:
Does rolling this buildroot change into flutter/engine have coupled engine changes? I'm not able to roll the buildroot past this commit without breaking Android builds.

shihaohong:
It does have coupled engine changes, since the android sdk and its related tools (build-tools, platform-tools, etc) need to be updated that have not been made it into the engine repo yet. What is the usual process when something like this occurs? It's my first time working on the buildroot and was originally wondering about this issue as well.

shihaohong:
One solution could be to update the DEPS file in the engine change to [match the SDK and tool version changes I was going to make](https://github.com/flutter/engine/pull/10413/files). That way, the Android build should be fine.

stuartmorgan:
For now I'm going to roll back instead of forward. This is only an issue because the patch I landed just before this one broke the tree.

If you can roll the buildroot forward to ToT tomorrow, that will unblock the Windows changes that depend on my buildroot patches.

PR #294, opened 2019-08-07
Revert "Add support for ANGLE using upstream build rules (#290)"

stuartmorgan:
Reverting due to ShellTest failures on Windows.


stuartmorgan:
Reverting on the buildroot side to avoid creating problems in other buildroot rolls until I can debug and resolve the ShellTest failures.

PR #296, opened 2019-08-07
Reland: Add support for ANGLE using upstream build rules

stuartmorgan:
This adds minimal (in some cases dummy) files sufficient to allow
building ANGLE in this buildroot using the build_with_chromium
option.

This relanded version uses the new swiftshader_ prefixed versions of the
DLL names in the swiftshader build, which are loaded preferentially if
available. This ensures that the swiftshader tests will use those,
rather than the potentially incompatible ANGLE versions. In local
testing, this resolved the ShellTest failures that led to the previous
revert.


stuartmorgan:
Once this lands I'll roll it into the engine right away so that I can roll it back if there are still problems, before it affects other buildroot PRs.

PR #310, opened 2019-09-05
Add new lines to eof

dnfield:


dnfield:
find . -name "*.cc" -print0 | xargs -0 -L1 bash -c 'test "$(tail -c 1 "$0")" && echo >> $0'

PR #324, opened 2019-10-08
Allow ninja on Windows to honor the fact that .lib-files are not touched if there are no changes.

aam:
Fixes https://github.com/flutter/flutter/issues/42021


liyuqian:
BTW, @chinmaygarde : should we setup some build tests in engine repo that check no-op build in any platform should be completed within 1 second? I think it may be as easy as to just modify our `.cirrus.yml` script, and probably add a very simple test script.

PR #330, opened 2019-10-21
Remove win "/WX" 'treat warnings as errors' flag from global flags.

GaryQian:
This will be instead added to flutter_root:config (in harfbuzz PR https://github.com/flutter/engine/pull/13242)

This rule causes harfbuzz to fail to compile on windows


GaryQian:
Yeah, I should take care of this already in https://github.com/flutter/engine/pull/13242/files#r337262189

PR #336, opened 2019-12-04
Fix LLDB can't work with --build-id

lizhangqu:
See [Why lldb remote debug can't locate the libflutter.so's symbol file](https://github.com/flutter/flutter/issues/45971)

See [Build System Maintainers Guide](https://android.googlesource.com/platform/ndk/+/refs/heads/ndk-release-r21/docs/BuildSystemMaintainers.md)
`Android Studio‘s LLDB debugger uses a binary’s build ID to locate debug information. To ensure that LLDB works with a binary, pass an option like -Wl,--build-id=sha1 to Clang when linking. Other --build-id= modes are OK, but avoid a plain --build-id argument when using LLD, because Android Studio‘s version of LLDB doesn’t recognize LLD's default 8-byte build ID. See Issue 885.`

See [[BUG] Breakpoints never hit in native code when using LLD](https://github.com/android/ndk/issues/885)

### DIFF
--- a/build/toolchain/custom/BUILD.gn
+++ b/build/toolchain/custom/BUILD.gn
@@ -81,7 +81,7 @@ toolchain("custom") {
     # existing .TOC file, overwrite it, otherwise, don't change it.
     tocfile = sofile + ".TOC"
     temporary_tocname = sofile + ".tmp"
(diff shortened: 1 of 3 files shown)


lizhangqu:
@googlebot I signed it!

cbracken:
Thank you for rebasing to head. Merging!

PR #338, opened 2019-12-13
Fix --format option for 'nm'

dwrobel:
llvm-nm doesn't seem to accept shorten form of --format arguments,
see below an error excerpt:
 arm-rdk-linux-gnueabi-llvm-nm: for the   --format option: Cannot find option named 'p'!

So, let's pass the full name of the option (which also works for binutils version of nm).

Signed-off-by: Damian Wrobel <dwrobel@ertelnet.rybnik.pl>


dwrobel:
@googlebot I signed it!

cbracken:
Thanks for the fix and apologies for the delay in getting it merged.

PR #345, opened 2020-02-14
Fix create_updated_flutter_deps to handle repos at different paths.

sjindel-google:
We can't rely on the repo name to identify what version tag it should use, so we use the folder name instead. Unfortunately the folder name is incorrect for "quiver-dart", so we special case it.

### DIFF
--- a/tools/dart/create_updated_flutter_deps.py
+++ b/tools/dart/create_updated_flutter_deps.py
@@ -111,9 +111,9 @@ def Main(argv):
                 updated_value = dart_v.replace(new_vars["dart_git"], "Var('dart_git') + '/")
                 updated_value = updated_value.replace(old_vars["chromium_git"], "Var('chromium_git') + '")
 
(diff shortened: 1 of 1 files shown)


tools/dart/create_updated_flutter_deps.py
```diff
@@ -111,9 +111,9 @@ def Main(argv):
                 updated_value = dart_v.replace(new_vars["dart_git"], "Var('dart_git') + '/")
                 updated_value = updated_value.replace(old_vars["chromium_git"], "Var('chromium_git') + '")
 
-                plain_v = dart_v[dart_v.rfind('/')+1:dart_v.rfind('@')]
-                if plain_v.rfind('.git') != -1:
-                  plain_v = plain_v[:plain_v.rfind('.git')]
+                plain_v = dart_k[dart_k.rfind('/') + 1:]
+                if plain_v == "quiver":
```
mkustermann: Please add a comment why this is special cased here, namely that the repository name is not the same as the directory name

PR #356, opened 2020-04-02
Use buildtools ar when building Android

dnfield:
See comment - the mistmatch between ar versions (in particular, the NDK's ar being as old as it is) causes linker failures later.

@jason-simmons @chinmaygarde

### DIFF
--- a/build/toolchain/android/BUILD.gn
+++ b/build/toolchain/android/BUILD.gn
@@ -65,7 +65,9 @@ template("android_gcc_toolchain") {
       cxx = compiler_prefix + tool_prefix + "g++"
     }
 
-    ar = tool_prefix + "ar"
+    # The NDK shipped ar utility is out of date, and using it will cause linker
+    # failures for version inconsistencies.
+    ar = "ar"
     ld = cxx
     readelf = compiler_prefix + tool_prefix + "readelf"
     nm = compiler_prefix + tool_prefix + "nm"
--- a/build/toolchain/android/BUILD.gn
+++ b/build/toolchain/android/BUILD.gn
@@ -60,14 +60,13 @@ template("android_gcc_toolchain") {
 
       cc = compiler_prefix + prefix + "/clang"
       cxx = compiler_prefix + prefix + "/clang++"
+      ar = prefix + "ar"
     } else {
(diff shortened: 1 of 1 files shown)


build/toolchain/android/BUILD.gn
```diff
@@ -65,7 +65,9 @@ template("android_gcc_toolchain") {
       cxx = compiler_prefix + tool_prefix + "g++"
     }
 
-    ar = tool_prefix + "ar"
+    # The NDK shipped ar utility is out of date, and using it will cause linker
+    # failures for version inconsistencies.
+    ar = "ar"
```
jason-simmons, line 70: Which version of `ar` will this map to?  Is it going to use the `ar` in the environment's `PATH`?

I tried doing an `android_release` build with Clang 11 using:
`rebase_path("//buildtools/$host_dir/clang/bin", root_build_dir) + "/llvm-ar"`

So far that seems to be working.

dnfield: This is how it's set up in the iOS toolchain - although yes, I think you're right.  I'll switch this to your suggestion so we're sure to get the right ar from buildtools.

build/toolchain/android/BUILD.gn
```diff
@@ -60,12 +60,13 @@ template("android_gcc_toolchain") {
 
       cc = compiler_prefix + prefix + "/clang"
       cxx = compiler_prefix + prefix + "/clang++"
+      ar = prefix + "/llvm-ar"
```
dnfield, line 63: @jason-simmons - I branched ar here so that if someone isn't using clang they fall back to the NDK ar.

build/toolchain/android/BUILD.gn
```diff
@@ -60,14 +60,13 @@ template("android_gcc_toolchain") {
 
       cc = compiler_prefix + prefix + "/clang"
       cxx = compiler_prefix + prefix + "/clang++"
+      ar = prefix + "ar"
```
jason-simmons: this should be `prefix + "/ar"`

dnfield: Fixed that - I commited before saving the file again :)

PR #357, opened 2020-04-03
Remove the Java compile time dependency on rt.jar

jason-simmons:
rt.jar is not provided by recent versions of the JDK

### DIFF
--- a/build/android/gyp/javac.py
+++ b/build/android/gyp/javac.py
@@ -92,21 +92,7 @@ def DoJavac(
       '-classpath', ':'.join(classpath),
       '-d', classes_dir]
 
-  def GetJavaHome():
-    # Android's boot jar doesn't contain all java 8 classes.
-    # See: https://github.com/evant/gradle-retrolambda/issues/23.
-    # Get the path of the jdk folder by searching for the 'jar' executable. We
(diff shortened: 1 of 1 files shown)


build/android/gyp/javac.py
```diff
@@ -92,21 +92,7 @@ def DoJavac(
...
-    # version of 'javac'.
-    # Can't use jar path for macOS though.
-    if sys.platform == 'darwin':
-      return subprocess.check_output(['/usr/libexec/java_home', '-v', '1.8']).strip()
-    jar_path = os.path.realpath(distutils.spawn.find_executable('jar'))
-    return os.path.dirname(os.path.dirname(jar_path))
-
   if bootclasspath:
-    rt_jar = os.path.join(GetJavaHome(), 'jre', 'lib', 'rt.jar')
-    bootclasspath.append(rt_jar)
```
dnfield, line 109: Does this still compile and run properly with your related engine patch?

jason-simmons: Yes - once both patches are approved I will add a buildroot roll to the engine patch

dnfield: The answer is yes.  Cool, TIL

PR #358, opened 2020-04-06
Update Chromium sysroot scripts

stuartmorgan:
Updates the very old, currently unused scripts for fetching Linux
sysroots to the latest version from the Chromium tree. This allows us to
start using the script to get a sysroot for our Linux builds. In the
future we may want to build our own, but for now the Chromium one seems
to work for us.

Also removes the old scripts for creating sysroots. If we decide to
create our own, we should start from the current Chromium scripts, or
potentially from scratch, but not from these old versions.

Part of https://github.com/flutter/flutter/issues/53176


stuartmorgan:
> I trust this is just a replacement of old stuff we did not author with new stuff (which also has been left as-is).

Correct; no changes.

PR #359, opened 2020-04-07
Use a sysroot for Linux host builds by default

stuartmorgan:
If no explicit sysroot is set, and it's not disabled, use a sysroot for
Linux host builds. This means we will no longer require specific
libraries to be on CI images for host builds, and developers won't need
as much local setup.

Fixes https://github.com/flutter/flutter/issues/53176

### DIFF
--- a/build/config/sysroot.gni
+++ b/build/config/sysroot.gni
@@ -10,6 +10,10 @@ declare_args() {
   # the target toolchain.
   target_sysroot = ""
 
+  # Whether to use the default sysroot when building for Linux, if an explicit
+  # sysroot isn't set.
+  use_default_linux_sysroot = true
+
   # The absolute path to the Xcode toolchain. This is used to look for headers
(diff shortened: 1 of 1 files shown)


stuartmorgan:
Do the conditions on this look right? The goal is for it to apply only to the Linux host build, not variants (e.g., Fuchsia) or non-host builds (e.g., Android).

I can't test this against the engine CI without landing it here first AFAICT, unfortunately, due to the split buildroot/engine repo situation.

stuartmorgan:
Actually, I should be able to do a canary PR against this branch, so I'll do that before landing here.

stuartmorgan:
Well, `is_fuchsia_host` is always false in our buildroot, so I assume that's something that it only makes sense to check in flutter/engine (which I assume uses a different buildroot for fuchsia host builds?).

Test PR passed compilation, so I'll go ahead and land this.

PR #360, opened 2020-04-08
Add shims for Chromium Khronos compatibility

stuartmorgan:
This makes the buildroot compatible with the Chromium
Khronos repo, allowing it to be added as a third_party dependency of the
engine.

Needed for https://github.com/flutter/flutter/issues/30729

### DIFF
--- a/gpu/GLES2/gl2chromium.h
+++ b/gpu/GLES2/gl2chromium.h
@@ -1,0 +1,5 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// No content needed. See //gpu/README.md.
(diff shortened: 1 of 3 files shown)


stuartmorgan:
Thoughts? It's a bit ugly, and I know we generally want our buildroot to be less of a semi-Chromium frankenstein rather than more, but it does let us trivially use [a repository](https://chromium.googlesource.com/chromium/src/third_party/khronos/) that has what we need.

Other options:
- Use the upstream Khronos repos (which are apparently in the hundreds of MB due to files we don't care about) plus a (trivial) `secondary` BUILD.gn
- Check the Khronos header into `flutter/engine`'s third_party; it seems like that's something that the team has generally not wanted to add to.
- Make our own repo like the Chromium one, but without the Chromium changes that require these shims (assuming we have the ability to do that; is there precedent? So far I'm just aware of us using Chromium and Fuchsia third_party repos)
- Build our own sysroot with these headers (assuming they are available as a package on Debian) and don't add Khronos to the repo at all.

PR #377, opened 2020-06-17
Delete a lot of unused files

dnfield:
These are all old, stale, way out of date copies of the Chromium buildroot.

There are definitely more files that could be removed, but I'm getting scared of what might fail on post submit, particularly for platforms I'm not as familiar with building.

There's a test run of rolling this here: https://github.com/flutter/engine/pull/19106

@goderbauer @chinmaygarde @stuartmorgan


build/ios/OWNERS
```diff
@@ -1,4 +1,2 @@
 rohitrao@chromium.org
 stuartmorgan@chromium.org
```
stuartmorgan, line 2: 🤣 

I'm pretty sure Rohit and I don't need to be owners of this directory in the Flutter buildroot, so you could just remove this file. (And, presumably, any other OWNERS files we have lying around, since (a) they will all be wrong, and (b) we don't use OWNERS AFAIK.)

jonahwilliams:
I didn't know there was this much red in the whole galaxy

build/ios/OWNERS
dnfield: deleted all owners files

goderbauer:
This is beautiful!

LGTM assuming that everything still builds :P

dnfield:
This has passed all presubmits. Hopefully it passes post. Going to land to start a roll.

PR #379, opened 2020-06-18
Remove .lib from link outputs on Windows.

aam:
Binaries are not exporting any symbols in flutter/engine settting, so no .lib files are generated.

Fixes https://github.com/flutter/flutter/issues/59687

### DIFF
--- a/build/toolchain/win/BUILD.gn
+++ b/build/toolchain/win/BUILD.gn
@@ -188,7 +188,6 @@ template("msvc_toolchain") {
       description = "LINK $binary_output"
       outputs = [
         binary_output,
-        "{{root_out_dir}}/{{target_output_name}}.lib",
       ]
 
       # The use of inputs_newline is to work around a fixed per-line buffer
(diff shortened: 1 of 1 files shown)


build/toolchain/win/BUILD.gn
```diff
@@ -188,7 +188,6 @@ template("msvc_toolchain") {
       description = "LINK $binary_output"
       outputs = [
         binary_output,
-        "{{root_out_dir}}/{{target_output_name}}.lib",
       ]
```
stuartmorgan, line 191: Could you add the pdb file here? It is an output of the build command, and the Chromium rules list it, so it seems like it should be here. (It probably doesn't matter much since presumably nothing is depending on that output, but we may as well be complete.)

eseidelGoogle: I spent a while reading the chromium build.gn and pdb integration I think will require more work.

I suggest we land this as-is and do pdb separately?  But I defer to you @stuartmorgan who are more active in this code than I. :)

eseidelGoogle: Again, I defer to you. I had never heard of pdb until yesterday. :)

When I briefly looked at copying some of Chromium's rules, pdb_output was mentioned several times in the file and I didn't look in detail as to how much work it would be to integrate it properly:
https://chromium.googlesource.com/chromium/src/+/5f1d704d671ca80af36a67c6899ab846b2397254/build/toolchain/win/BUILD.gn

PR #385, opened 2020-07-28
Update third_party/zlib from Chromium

stuartmorgan:
Replaces third_party/zlib, which was originally forked from Chromium,
with the latest copy from Chromium.

BUILD.gn contains some changes to build in our buildroot without
cascading changes:
- Removes the `build_with_chromium` sections.
- Removes optimization configuration adjustments that rely on
  buildroot-wide configuration options that are newer than our fork
  (removing default optimization settings, max_speed)
- Removes an is_nacl condition, adds a previously-locally-added
  is_fuchsia check.

This is needed to allow the Chromium zlib additions added in
https://github.com/flutter/buildroot/pull/373 to build correctly.


stuartmorgan:
I'm not sure how I missed that the additional files I added before had a compile error against our older copy of zlib.

Rather than continue to cherry-pick bits and pieces, it seemed better to just pull a full updated copy from Chromium.

stuartmorgan:
Looks like this will require some non-trivial license updates (i.e., script, not just file path lists) so I'm making a separate roll PR for it (https://github.com/flutter/engine/pull/20119) rather than rolling it as part of the ANGLE roll PR as originally planned.

PR #388, opened 2020-07-29
Differentiate clang invocation when targeting x86 vs x64 on Windows.

aam:
This fixes building of android gen_snapshot with clang.

This is along the same lines that were done in dart sdk https://dart-review.googlesource.com/c/sdk/+/142704

### DIFF
--- a/build/toolchain/win/BUILD.gn
+++ b/build/toolchain/win/BUILD.gn
@@ -78,7 +78,7 @@ template("msvc_toolchain") {
       #precompiled_header_type = "msvc"
       pdbname = "{{target_out_dir}}/{{target_output_name}}_c.pdb"
       flags = ""
-      if (invoker.current_cpu == "x86") {
+      if (is_clang && invoker.current_cpu == "x86") {
         flags = "-m32"
       }
       command = "ninja -t msvc -e $env -- $cl $flags /nologo /showIncludes /FC @$rspfile /c {{source}} /Fo{{output}} /Fd$pdbname"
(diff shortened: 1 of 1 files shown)


stuartmorgan:
Great! I hadn't had time to come back and investigate the issues with using clang for the Android Window host build yet.

Is there a corresponding PR already to enable clang for that configuration in the GN wrapper script?

build/toolchain/win/BUILD.gn
```diff
@@ -77,7 +77,11 @@ template("msvc_toolchain") {
       # TODO(brettw) enable this when GN support in the binary has been rolled.
       #precompiled_header_type = "msvc"
       pdbname = "{{target_out_dir}}/{{target_output_name}}_c.pdb"
-      command = "ninja -t msvc -e $env -- $cl /nologo /showIncludes /FC @$rspfile /c {{source}} /Fo{{output}} /Fd$pdbname"
+      flags = ""
+      if (is_clang && invoker.current_cpu == "x86") {
+        flags = "-m32"
```
stuartmorgan, line 82: Chromium's approach is to add this [in build/config/win](https://source.chromium.org/chromium/chromium/src/+/master:build/config/win/BUILD.gn;l=111?q=m32%20file:gn$&ss=chromium&originalUrl=https:%2F%2Fcs.chromium.org%2F); I have a local change where I was experimenting with Windows arm64 support where I ended up needing to add that block as well. I can move this whenever I land the initial arm plumbing.

aam:
> Great! I hadn't had time to come back and investigate the issues with using clang for the Android Window host build yet.
> 
> Is there a corresponding PR already to enable clang for that configuration in the GN wrapper script?

Yes, when this change was rolled into the engine, clang was enabled for android(on Windows).

build/toolchain/win/BUILD.gn
aam: Good thing about current setup is that it is aligned with dart https://github.com/dart-lang/sdk/blob/master/build/toolchain/win/BUILD.gn.
What are the advantages of Chromiums approach?

stuartmorgan: > is that it is aligned with dart

Since I've routinely needed to re-sync and/or cherry-pick things from Chromium when adding new functionality (like arm64 support) to the Flutter build system, I don't know that "aligned with Dart" is more of an advantage than "aligned with Chromium".

> What are the advantages of Chromiums approach?

I don't have an authoritative answer, but personally this looks a lot to me like partially re-inventing `cflags`, which is already a built-in concept.

PR #391, opened 2020-08-07
Forward the frameworks variable in the component template

jason-simmons:


build/config/BUILDCONFIG.gn
```diff
@@ -741,6 +741,9 @@ template("component") {
       if (defined(invoker.forward_dependent_configs_from)) {
         forward_dependent_configs_from = invoker.forward_dependent_configs_from
       }
+      if (defined(invoker.frameworks)) {
```
zanderso, line 744: Is this also needed under the `shared_library()` target for when `is_component_build` is true?

jason-simmons: added it there for consistency

PR #392, opened 2020-08-11
Switch to VS 2019

stuartmorgan:
This is a reland of https://github.com/flutter/buildroot/pull/339


stuartmorgan:
The blocking issues that led to the revert have been resolved:
- ANGLE has been rolled to a new enough version to be 2019-compatible
- The last of the Windows bots have moved to clang, so we don't have goma incompatibilities.

The last run of https://github.com/flutter/engine/pull/18045 looks good (other than an unrelated Dart analysis failure).

PR #393, opened 2020-08-13
account for macos 11 too when looking for mac sdks

xster:
Fixes https://github.com/flutter/flutter/issues/63613

Ran into it to debug https://github.com/flutter/flutter/issues/60013

### DIFF
--- a/build/mac/find_sdk.py
+++ b/build/mac/find_sdk.py
@@ -55,7 +55,7 @@ def main():
     sdk_dir = xcode43_sdk_path
   else:
     sdk_dir = os.path.join(out.rstrip(), 'SDKs')
-  sdks = [re.findall('^MacOSX(10\.\d+)\.sdk$', s) for s in os.listdir(sdk_dir)]
+  sdks = [re.findall('^MacOSX(1[0|1]\.\d+)\.sdk$', s) for s in os.listdir(sdk_dir)]
   sdks = [s[0] for s in sdks if s]  # [['10.5'], ['10.6']] => ['10.5', '10.6']
   sdks = [s for s in sdks  # ['10.5', '10.6'] => ['10.6']
           if parse_version(s) >= parse_version(min_sdk_version)]


build/mac/find_sdk.py
```diff
@@ -55,7 +55,7 @@ def main():
     sdk_dir = xcode43_sdk_path
   else:
     sdk_dir = os.path.join(out.rstrip(), 'SDKs')
-  sdks = [re.findall('^MacOSX(10\.\d+)\.sdk$', s) for s in os.listdir(sdk_dir)]
+  sdks = [re.findall('^MacOSX(1[0|1]\.\d+)\.sdk$', s) for s in os.listdir(sdk_dir)]
```
jmagman, line 58: Isn't it MacOSX10.16.sdk?

xster: no, we're going for 11 https://en.wikipedia.org/wiki/MacOS_Big_Sur :D

jmagman: But the SDK is 10.16 https://appleinsider.com/inside/macos-10.16
```
drwxr-xr-x  5 magder  primarygroup  160 Jun 29 16:01 DriverKit20.0.sdk
drwxr-xr-x  7 magder  primarygroup  224 Jun 29 16:01 MacOSX.sdk
lrwxr-xr-x  1 magder  primarygroup   10 Jul  8 13:45 MacOSX10.16.sdk -> MacOSX.sdk
```

xster: huh, interesting. There's some nuanced stuff going on... I'm getting 

```
ll /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/                                                                                                                
total 0
drwxr-xr-x  5 xster  staff   160B Jul 30 05:11 DriverKit20.0.sdk/
drwxr-xr-x  7 xster  staff   224B Jul 30 05:11 MacOSX.sdk/
lrwxr-xr-x  1 xster  staff    10B Aug 11 13:11 MacOSX11.0.sdk@ -> MacOSX.sdk
```

Which beta version you on?

xster: can't find any other discussions about this... https://developer.apple.com/documentation/xcode-release-notes/xcode-12-beta-release-notes says `Xcode 12 beta 4 includes SDKs for iOS 14, iPadOS 14, tvOS 14, watchOS 7, and macOS 11` and nothing for beta 3. But `macOS 11` here could also just be referencing the marketing term...

Would love to see what you get on Xcode 12 beta 4.

jmagman: I was on beta 2.  You're right, it's MacOSX11.0.sdk on beta 4!

xster: ah cool, good to know. Either way, this should handle 10.16 or 11.0.

jmagman: Can we uses `^MacOSX\d+.*\.sdk$` so we don't need to do this again next year?  It looks like it's trying to not match `MacOSX.sdk`

xster: Let's be conservative. I think it's not a bad thing for someone to come back and potentially touch this file again in case macos does decide to keep going up a major version every year.

PR #395, opened 2020-08-13
Upgrade android sdk version to 30r2

GaryQian:
This updates the buildroot to use Android 30r2 and build-tools 30.0.1.

Rolls into engine with https://github.com/flutter/engine/pull/20479

PR #396, opened 2020-08-15
Revert "Switch to VS 2019"

GaryQian:
Reverts flutter/buildroot#392

Temporarily revert so that changes in https://github.com/flutter/buildroot/pull/395 can be rolled.

This seems to cause https://ci.chromium.org/p/flutter/builders/try/Windows%20Host%20Engine/11969? which is part of https://github.com/flutter/engine/pull/20479


stuartmorgan:
https://github.com/flutter/engine/pull/18045 is my ongoing attempt to roll that buildroot change into the engine, and it's definitely broken. I thought I had one pass before landing 392 that worked, but looking back it looks like I missed the Windows exception and thought it was just the unrelated Linux failure.

PR #399, opened 2020-09-02
OCMock for Mac tests

justinmc:
OCMock is currently set up only for iOS, but I need it in a Mac test in https://github.com/flutter/engine/pull/20531.

PR #402, opened 2020-09-11
Fix OCMock static_library target

cbracken:
In flutter/buildroot#399, the static_library target was split into a
source_set and static_library target, but this resulted in libocmock
being built as an empty 88-byte library. We never caught the issue since
this buildroot change was never rolled to the engine.

This reverts things back to match how we build other static libraries
such as libzip and libxml.

This reverts 940dcf6e85301c049576da58b193cf2d903539a2.

### DIFF
--- a/build/secondary/third_party/ocmock/BUILD.gn
+++ b/build/secondary/third_party/ocmock/BUILD.gn
@@ -9,13 +9,18 @@ config("ocmock_config") {
   include_dirs = [ "$ocmock_path" ]
 }
 
-source_set("ocmock_src") {
+static_library("ocmock") {
   configs -= [ "//build/config/compiler:chromium_code" ]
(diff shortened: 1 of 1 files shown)


xster (approved):

justinmc:
This will break my PR that depends on ocmock_src. See https://github.com/flutter/engine/pull/20531/files#diff-934b1a4a722f53c991b1f4445b307d3cR120

I needed a separate way to include ocmock for Mac and for iOS.

cbracken:
The way the buildroot change was written was resulting in an empty static_library with no symbols (i.e. it was broken at head). I can take a look at relanding that so it generates a static_library with content.

justinmc:
Alright thanks. I'm out of my usual area of expertise, but let me know if there's anything I can take a look at myself if it's not trivial for you.

PR #403, opened 2020-09-11
Re-split ocmock into source set and static lib

cbracken:
This re-lands #399 with complete_static_lib=true in the static library
target.

static_library does not includes sources from its dependencies, and
since this target had no (direct) sources of its own, was being built
empty in the original version of this split. The complete_static_lib
forces the library to include sources from its dependencies.

### DIFF
--- a/build/secondary/third_party/ocmock/BUILD.gn
+++ b/build/secondary/third_party/ocmock/BUILD.gn
@@ -9,18 +9,15 @@ config("ocmock_config") {
   include_dirs = [ "$ocmock_path" ]
 }
 
-static_library("ocmock") {
+# Target that compiles all sources to .o files but does not produce a static
+# library for use in macOS desktop tests.
+source_set("ocmock_src") {
   configs -= [ "//build/config/compiler:chromium_code" ]
   all_dependent_configs = [ ":ocmock_config" ]
   cflags = [
     "-fvisibility=default",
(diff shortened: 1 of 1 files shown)


gaaclarke (approved):

cbracken:
Context is that this is required for https://github.com/flutter/engine/pull/20531, which I don't have the context on. @justinmc  I'm curious why we're unable to link against the static library directly?

justinmc:
I was running into an error described in my [comment here](https://github.com/flutter/engine/pull/20531#issuecomment-685882845) when I tried to do this directly.  CC @gaaclarke who helped me overcome that.

With this PR, I now get a new error:

```
$ ninja -C out/host_debug_unopt -j 100
ninja: Entering directory `out/host_debug_unopt'
ninja: error: '../../third_party/ocmock/Source/OCMock/OCMNonRetainingObjectReturnValueProvider.m', needed by 'obj/third_party/ocmock/Source/OCMock/ocmock_src.OCMNonRetainingObjectReturnValueProvider.o', missing and no known rule to make it
```

cbracken:
For archaeologists `gclient sync` should fix the above.

justinmc:
Just confirmed that indeed `gclient sync` fixed it for me 👍

justinmc (approved):
LGTM

PR #404, opened 2020-10-12
Create a dylib ocmock target

dnfield:
This will help us clean up some duplicate symbol errors, and close out flutter/flutter#54504


gaaclarke (approved):
I just looked at this friday.  I think this should solve the problem.  LGTM.

PR #409, opened 2020-10-31
Add UWP Config

clarkezone:
Per code review feedback from https://github.com/flutter/engine/pull/21754 moving the flutter_windows_winuwp_defs config to buildroot

### DIFF
--- a/build/config/win/BUILD.gn
+++ b/build/config/win/BUILD.gn
@@ -21,6 +21,10 @@ config("sdk") {
     # This is required for ATL to use XP-safe versions of its functions.
     "_USING_V110_SDK71_",
   ]
+
+  if (target_os == "winuwp") {
+    defines += [ "WINUWP" ]
+  }
 }
 
 # Sets the default Windows build version. This is separated because some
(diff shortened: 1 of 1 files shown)


build/config/win/BUILD.gn
```diff
@@ -32,6 +32,11 @@ config("winver") {
   ]
 }
 
+# When OS target is WinUWP use WINUWP as the preprocessor flag to conditionalize.
+config("flutter_windows_winuwp_defs") {
+  defines = [ "WINUWP" ]
```
stuartmorgan, line 37: Shouldn't we follow [the Chromium approach](https://source.chromium.org/chromium/chromium/src/+/master:build/config/win/BUILD.gn;l=223?q=defines%20winuwp&ss=chromium) and put this in the `sdk` config above behind an OS check, so that it's automatically applied to all the targets that are getting the other defines?

clarkezone: Great idea.  Addressed.

stuartmorgan (approved):
LGTM

PR #407, opened 2020-10-28
Update buildroot for ANGLE roll

stuartmorgan:
Updates build/ to match the current expectations of an angle_has_build
build.

PR #410, opened 2020-11-04
Reland "Update buildroot for ANGLE roll"

stuartmorgan:
Relands https://github.com/flutter/buildroot/pull/407

Updates build/ to match the current expectations of an angle_has_build
build.

### DIFF
--- a/build/config/BUILDCONFIG.gn
+++ b/build/config/BUILDCONFIG.gn
@@ -284,6 +284,11 @@ if (current_os == "win") {
   is_win = false
 }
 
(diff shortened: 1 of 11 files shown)


stuartmorgan:
I have the ANGLE roll PR down to just needing to update the actual license files (license script updates and upstream license block fixes are all in place), which I'm about to run, so it's safe to move forward with this again.

stuartmorgan:
License check has now passed for the ANGLE roll PR, so relanding to move forward with that PR.

PR #390, opened 2020-08-05
Add support of cross-building  the engine for ARM64 Linux Host

HidenoriMatsubayashi:
## Description

Added cross-building Flutter Engine for ARM64 Linux platforms. This PR is part of ARM64 Linux support in the Flutter SDK.

## Related PRs
 - [ ] https://github.com/flutter/flutter/pull/61221
 - [x] https://github.com/flutter/engine/pull/20254

## Related Issues
* https://github.com/flutter/flutter/issues/56992
* https://github.com/flutter/flutter/issues/60678

### 1. Build Test
First, I confirmed that Flutter Engine can be built for x64 and arm64. The procedure is as follows. All are operations on x64 Linux Host.

Get source code and setup build enviroment.

```
create .gclient file.
$ gclient sync
$ cd src
````

Cross-build for ARM64 Linux

```
$ ./flutter/tools/gn --target-os linux --linux-cpu arm64 --runtime-mode release
$ ninja -C out/linux_release_arm64
-> Passed!

$ ./flutter/tools/gn --target-os linux --linux-cpu arm64 --runtime-mode debug
$ ninja -C out/linux_debug_arm64
-> Passed!
```

Build for x64 Linux and Regression test.

```
$ ./flutter/tools/gn --unoptimized
$ ninja -C out/host_debug_unopt
-> Passed!

$ ./flutter/testing/run_tests.sh
-> All Tests Passed!
```

### 2. Run Flutter SDK and app on ARM64 Linux (Jetson Nano)

Second, using the artifacts for ARM64 built by the above procedure, I have confirmed that the Flutter SDK and Flutter app can work fine on Jetson Nano (ARM64 CPU).

All are operations on ARM64 Linux Host (Jetson Nano).

```
$ git clone https://github.com/flutter/flutter.git
$ export PATH=$PATH:{your working dir path}/flutter/bin
$ flutter doctor
   ... Error happened.
```

Copy `dart-sdk` which was locally built to `${your installpath}/bin/cache/`.

```
$ flutter doctor
$ flutter config --enable-linux-desktop
$ flutter create sample
$ cd sample
$ flutter build linux
   ... Error happened.
```

Copy the Flutter SDK artifacts (build data) to `${your installpath}/bin/cache/artifacts/engine` directory manually.

 -  `out/linux_debug_arm64/flutter_patched_sdk` to `${your installpath}/bin/cache/artifacts/engine/common/flutter_patched_sdk`
[...]

PR #412, opened 2020-11-13
Fix toolchain for linux x86_64

HidenoriMatsubayashi:
## Background

In the `clang_x86` and `clang_x64` building, `buildroot` has been using readelf/nm/ar/strip commands of a host that executes the build without using the toolchain's. However, it is odd to use gcc in the host environment even though it uses clang / llvm.

> It looks like Dart is using llvm-objcopy for clang_x64 as well, so we should investigate whether to unify that at some point on our side, but limiting the changes to just arm64 here is the safer option.

https://github.com/flutter/buildroot/pull/390#pullrequestreview-522776808

## Changes

I investigated the past commit history of dart-sdk and flutter. I guess that the cause is that the changes of dart-sdk are not imported to flutter-engine.

dart-sdk was changed from gcc to clang/llvm by [this PR](https://github.com/dart-lang/sdk/commit/81e428fd39032fb23997c04c6681f8e5fcda9105#) on 21 May 2017. However, there isn't any changes in flutter. I don't think it's the critical problem, but I think it's better to modify the toolchain according to dart-sdk.

### clang_x64 and clang_x86

Changed readelf, nm, ar, strip(llvm_objcopy) commands from gcc to clang/llvm. The source code is the same with dart-sdk's.
https://github.com/dart-lang/sdk/blob/master/build/toolchain/linux/BUILD.gn

### x64 and x86
Formally, there isn't any changes without adding strip command. I just changed the code according to dart-sdk.

## Test
The tests in the CI passed.
https://github.com/flutter/engine/pull/22484

### DIFF
--- a/build/toolchain/linux/BUILD.gn
+++ b/build/toolchain/linux/BUILD.gn
@@ -40,6 +40,7 @@ gcc_toolchain("clang_x86") {
   }
   cc = "${compiler_prefix}$prefix/clang"
   cxx = "${compiler_prefix}$prefix/clang++"
+
   readelf = "${prefix}/llvm-readelf"
   nm = "${prefix}/llvm-nm"
   ar = "${prefix}/llvm-ar"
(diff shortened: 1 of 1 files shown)


dnfield (approved):
LGTM

HidenoriMatsubayashi:
Thank you for your review.

PR #417, opened 2020-12-11
Make symlinks so that RBE will work for Xcode toolchains

dnfield:
See https://github.com/flutter/flutter/issues/72032

This creates symlinks for the macOS SDK, iPhone SDK, and Xcode toolchain sysroot in the gen dir. That way, RBE won't fail when it tries to access a file based on an absolute path to /Applications/whatever on mac based host/iOS builds.

### DIFF
--- a/build/mac/find_sdk.py
+++ b/build/mac/find_sdk.py
@@ -10,13 +10,23 @@ Usage:
   python find_sdk.py 10.6  # Ignores SDKs < 10.6
 """
 
+import errno
 import os
 import re
 import subprocess
 import sys
 
-
 from optparse import OptionParser
+
+
+def mkdir_p(path):
+  try:
+    os.makedirs(path)
+  except OSError as exc:
+    if exc.errno == errno.EEXIST and os.path.isdir(path):
+      pass
+    else:
+      raise
 
 
 def parse_version(version_str):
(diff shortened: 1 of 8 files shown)


zanderso:
Do the symlinks need to go in a .gitignore?

dnfield:
They're created in a generated out directory - e.g. `out/host_debug_unopt/gen/MacOSX11.0.sdk`.

build/mac/find_sdk.py
```diff
@@ -10,15 +10,25 @@
...
 import os
 import re
 import subprocess
 import sys
 
-
 from optparse import OptionParser
 
 
+def mkdir_p(path):
```
zanderso, line 22: Is there any way to avoid copying this to several places?

dnfield: Yeah. I also realize that I need to fix this so it replaces the symlink incase you e.g. update or move Xcode for some reason.

build/pyutil/file_util.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2012 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import errno
+import os
+
```
zanderso, line 7: nit: please add comments

zanderso (approved):

build/pyutil/file_util.py
dnfield: Done

PR #420, opened 2020-12-16
Suppress warnings for fuchsia toolchain

chaselatta:
This change suppresses warnings that are introduced when
updating the fuchsia toolchain to clang 12.


arbreng (approved):

zanderso (commented):
lgtm

PR #426, opened 2021-01-13
Reland - prepare to roll fuchsia to clang 12

chaselatta:


zanderso (approved):

zanderso:
Do we need both this and https://github.com/flutter/buildroot/pull/427 ?

chaselatta:
We probably only need Dan's change, I am rolling clang for fuchsia only which is why I added this but if we are ready to land his PR then lets do that and I will abandon this one.

dnfield:
I'm working on just getting the whole repo on 12. I don't think landing both will break anything if it unblocks @chaselatta (I can always just update my patch to undo the `if is fuchsia` parts of this), but I expect to be able to get us on Clang 12 sooner than later.

dnfield:
It should be safe to land #427 even without rolling to clang 12 AFAICT though, because we're currently falling back to the system `as` anyway when building Android.

arbreng (approved):

chaselatta:
I am going to land this CL since rolling the toolchain is a current blocker for us. Feel free to remove the is_fuchsia check in the next PR.

PR #428, opened 2021-01-16
Roll zlib from Chromium's third_party/zlib repository

jason-simmons:


dnfield (approved):
LGTM.

I'd like to see this pulled in as a proper gclient dep but we can do that as a follow up.

dnfield:
Also once this rolls in we should be able to roll clang further forward.

PR #431, opened 2021-01-27
Fix visibility of symbols of libcxxabi

huanghongxun:
This time we cannot throw and catch C++ exceptions in plugin platform code. Because Flutter statically links libcxx and libcxxabi which exposes symbols like __cxa_decrement_exception_refcount, which will introduce two versions of cxxabi functions, throwing libcxx exceptions, and catching them by libstdc++ method.
To fix this problem, this PR set `__LIBCXXABI_DISABLE_VISIBILITY_ANNOTATIONS` to suppress symbol leaks, which is tested and worked correctly. All platforms artifacts will get benefit from this pr.

This pull request fixes issue flutter/flutter#66575 and flutter/flutter#74599.


dcharkes:
Verified locally that this solves flutter/flutter#74599.

stuartmorgan (approved):
LGTM, thanks for the PR!

PR #432, opened 2021-01-28
Add an include to expose use_xcode_clang to ANGLE

stuartmorgan:
ANGLE at ToT no longer sees the `use_xcode_clang` entry I added previously to satisfy its build (presumably due to some change in .gni includes in ANGLE). This pulls the .gni that defines it into the ANGLE-specific gni.

This is needed in order to roll ANGLE.


clarkezone (approved):
LGTM

PR #441, opened 2021-03-05
fuchsia: Update Vulkan headers.

gnoliyil:
This updates the Vulkan header version to 1.2.148
which Fuchsia currently uses.


gnoliyil:
@chinmaygarde

chinmaygarde (approved):
Clarified that this is just the updated contents of the Fuchsia `//third_party/Vulkan-Headers` with no modifications.

PR #442, opened 2021-03-08
Add script to generate winrt headers using cppwinrt.exe

clarkezone:
Generate winrt headers using the cppwinrt tool.  A subsequent engine PR will consume them.

https://github.com/flutter/flutter/issues/70205

Consumed by PR:
https://github.com/flutter/engine/pull/24861

### DIFF
--- a/build/win/generate_winrt_headers.py
+++ b/build/win/generate_winrt_headers.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+# Copyright 2014 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import os
+import sys
+
+def clean(output_dir):
+  if os.path.exists(output_dir):
+    os.remove(output_dir)
+  return
+
+
+def generate_headers(output_dir):
+  """Run cppwinrt.exe on the installed Windows SDK version and generate
+  cppwinrt headers in the output directory.
+  """
+  
+  cppwinrt_exe = os.path.join(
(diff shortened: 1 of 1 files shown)


build/win/generate_winrt_headers.py
```diff
@@ -0,0 +1,38 @@
...
+  """
+  
+  cppwinrt_exe = os.path.join(
+  __file__,
+  '..\\..\\third_party\\cppwinrt\\bin\\cppwinrt.exe')
+
+  args = [cppwinrt_exe, '-in', 'sdk',
+      '-out', '%s' % output_dir]
+
+  return 0
```
cbracken, line 27: Looks like we're not currently running the command here. Does this need a re-push?

cbracken (changes requested):

build/win/generate_winrt_headers.py
cbracken: Might also be a good idea for us to first recursively delete the output directory (if it exists) then run the command.

clarkezone: Good catch ;-)  I think i fat fingered the porting from my test branch.  On the delete front, I have the Clean() method that is called before `generate_headers()` but that isn't do a recursive.  I'll take another look to make sure that's working as expected

clarkezone: Both fixed in most recent commit

chinmaygarde (approved):

cbracken (commented):
Thanks!

PR #447, opened 2021-05-06
Make cppwinrt check SDK/References folder if using standard sdk location failed.

aam:


cbracken (approved):
This lgtm, but I'm curious why the SDK path doesn't match what's expected; once we know more we should update this script with a bit more info on why it's necessary.

PR #448, opened 2021-05-07
Use CIPD Windows SDK to generate WinRT headers

cbracken:
Rather than relying on the Windows SDK that happens to be on the host
machine, we now pull down the Windows SDK from CIPD into
third_party/windows_sdk and use it directly.

cppwinrt relies on a registry entry to determine the SDK location and
doesn't currently support a mechanism for specifying this location
explicitly via an argument. Instead, we do what cppwinrt itself does and
parse out the Platform.XML file to determine the set of .winmd files to
take as input, then pass them to the tool directly.

### DIFF
--- a/build/win/generate_winrt_headers.py
+++ b/build/win/generate_winrt_headers.py
@@ -7,7 +7,13 @@ import os
 import shutil
 import subprocess
 import sys
-import winreg
+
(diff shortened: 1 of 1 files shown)


clarkezone (approved):
LGTM

PR #449, opened 2021-05-07
Canonicalize cppwinrt input paths

cbracken:
Canonicalizes input paths we pass to cppwinrt to avoid failures to open
files.

Followup to https://github.com/flutter/buildroot/pull/448


clarkezone (approved):
LGTM

PR #450, opened 2021-05-07
Use an options file for cppwinrt

cbracken:
cppwinrt offers the ability to read a file containing its command-line
options rather than specifying a massive arguments list on the
command-line. This also enables easier debugging.

Also enables logging verbose output.


dnfield (approved):
LGTM

PR #451, opened 2021-05-07
Emit cppwinrt options file on failure

cbracken:
cppwinrt emits the file contents (formatted slightly better) itself when
it passes.


clarkezone (approved):
LGTM

PR #460, opened 2021-06-04
Don't use -flto when building for host on mac

zanderso:
Linking `gen_snapshot` for the host with LTO takes about 3+ minutes, and probably doesn't provide a lot of benefit.


chinmaygarde (approved):

dnfield:
I had looked into doing this in the past, but decided against it because gen_snapshot is already slow and LTO is supposed to help speed it up. 

That said, I never benchmarked it with and without this - but if we see compilation regressions on this roll we should probably suspect this PR.

PR #461, opened 2021-06-04
Update sysroot_ld_path.py for Python3

cbracken:
In Python 3, subprocess.check_output returns an encoding-independent
byte array. If passed universal_newlines=True, both Python 2 and 3
decode and return a UTF-8 string.

Issue: https://github.com/flutter/flutter/issues/83043


chinmaygarde (approved):

cbracken:
I noticed that Python 2 appears to have been removed from my gLinux machine overnight. Eons ago I signed up to be on the testing channel, so perhaps this is finally on the way?

PR #471, opened 2021-06-17
Support iOS ARM64 simulator builds

gw280:
Fixes https://github.com/flutter/flutter/issues/84670

This was previously landed and backed out for this failure: https://ci.chromium.org/ui/p/flutter/builders/prod/Mac%20iOS%20Engine%20Release/7004/overview

I have done a full CI pre-submit run here: https://github.com/flutter/engine/pull/26784 and manually triggered a `Mac iOS Engine Release` run here: https://luci-milo.appspot.com/raw/build/logs.chromium.org/flutter/led/wrightgeorge_google.com/714c2f7d1c809fcbdd78d87e13fe2665d166438b0676ac67ea88cde200e6ff6e/+/build.proto

Both are passing (with the exception of failing to upload artifacts, due to being run on a try job), so I think we can merge this now.

### DIFF
--- a/build/config/BUILDCONFIG.gn
+++ b/build/config/BUILDCONFIG.gn
@@ -543,7 +543,11 @@ if (custom_toolchain != "") {
   import("//build/config/ios/ios_sdk.gni")  # For use_ios_simulator
   host_toolchain = "//build/toolchain/mac:clang_$host_cpu"
   if (use_ios_simulator) {
-    set_default_toolchain("//build/toolchain/mac:ios_clang_x64")
+    if (target_cpu == "arm" || target_cpu == "arm64") {
+      set_default_toolchain("//build/toolchain/mac:ios_clang_arm_sim")
+    } else {
+      set_default_toolchain("//build/toolchain/mac:ios_clang_x64_sim")
+    }
   } else {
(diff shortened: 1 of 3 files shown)


build/config/BUILDCONFIG.gn
```diff
@@ -543,7 +543,11 @@ if (custom_toolchain != "") {
   import("//build/config/ios/ios_sdk.gni")  # For use_ios_simulator
   host_toolchain = "//build/toolchain/mac:clang_$host_cpu"
   if (use_ios_simulator) {
-    set_default_toolchain("//build/toolchain/mac:ios_clang_x64")
+    if (target_cpu == "arm" || target_cpu == "arm64") {
```
cbracken, line 546: Does this work for armv7 builds? IIRC they use soft-fp ABI.

gw280: simulator doesn't actually use armv7, as the only possible hosts are ARM64 (on M1) or x86_64. I can remove the arm condition.

cbracken (approved):
lgtm if you've verified this works with armv7 builds.

build/config/BUILDCONFIG.gn
cbracken: sgtm -- that would make it clearer for readers.

PR #473, opened 2021-06-24
make fidl_gen_cpp work with python3

chaselatta:
This script is failing with python3. The code will attempt to call the python3 api and fallback to the python2 version if it fails.


cbracken (approved):

PR #512, opened 2021-09-16
Add clangd .cache to gitignore

arbreng:
This helps out a lot when using the vscode clangd extension


chinmaygarde (approved):
Looks fine. I just add this to my global .gitignore.

PR #517, opened 2021-10-12
Pass -fno-aligned-allocation for mac/arm64

zanderso:
We currently set the deployment target to 10.11, and aligned allocation is only supported on 10.14 and newer.


jmagman (approved):
LGTM

PR #526, opened 2021-11-18
Build script for the expat XML parser

jason-simmons:
See https://github.com/flutter/flutter/issues/91384


zanderso (approved):

zanderso:
Note for posterity: These look like the analogous files that are in the Skia repo.

PR #542, opened 2022-01-11
[fuchsia] Fix zx fidl bindings conflicting w/ SDK.

akbiggs:
The ZX FIDL library is special. Unlike regular FIDL libraries
which define IPC, they define the interface between
Zircon kernel and Fuchsia userspace. fidlgen
doesn't produce useful output when dealing with this
library. See fxb/90838.

We're seeing a conflict between zx_rights_t from
the SDK and zx_rights_t from ZX's FIDL bindings. To
fix this, we apply a similar change to https://fuchsia-review.googlesource.com/c/fuchsia/+/623061/13/scripts/sdk/gn/base/build/fidl_library.gni#192.

### DIFF
--- a/build/fuchsia/sdk.gni
+++ b/build/fuchsia/sdk.gni
@@ -101,10 +101,15 @@ template("_fuchsia_fidl_library") {
     ]
 
     outputs = [
-      "$target_gen_dir/$library_name_slashes/cpp/fidl.h",
-      "$target_gen_dir/$library_name_slashes/cpp/fidl.cc",
       "$target_gen_dir/$library_name_slashes/cpp/tables.c",
     ]
+    # TODO(https://fxbug.dev/90838): Make zx library less special-cased.
+    if (library_name != "zx") {
+      outputs += [
+        "$target_gen_dir/$library_name_slashes/cpp/fidl.h",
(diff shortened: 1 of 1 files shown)


build/fuchsia/sdk.gni
```diff
@@ -101,10 +101,15 @@ template("_fuchsia_fidl_library") {
     ]
 
     outputs = [
-      "$target_gen_dir/$library_name_slashes/cpp/fidl.h",
-      "$target_gen_dir/$library_name_slashes/cpp/fidl.cc",
       "$target_gen_dir/$library_name_slashes/cpp/tables.c",
```
arbreng, line 104: Do you still want to include tables.c for this library?

arbreng (approved):
LGTM, too bad we need this special casing....that is very non-obvious

build/fuchsia/sdk.gni
akbiggs: Yes based on https://fuchsia-review.googlesource.com/c/fuchsia/+/623061/13/scripts/sdk/gn/base/build/fidl_library.gni#192, but I'm not really sure why. Will check with Ian.

PR #543, opened 2022-01-13
Don't generate zx fidl headers and source.

chandarrengoog:
This is necessary to avoid name collisions.

See https://github.com/flutter/buildroot/pull/542.

### DIFF
--- a/build/fuchsia/fidl_gen_cpp.py
+++ b/build/fuchsia/fidl_gen_cpp.py
@@ -35,18 +35,19 @@ def main():
   parser = argparse.ArgumentParser();
 
   parser.add_argument('--fidlc-bin', dest='fidlc_bin', action='store', required=True)
-  parser.add_argument('--fidlgen-bin', dest='fidlgen_bin', action='store', required=True)
+  parser.add_argument('--fidlgen-bin', dest='fidlgen_bin', action='store', required=False)
 
   parser.add_argument('--sdk-base', dest='sdk_base', action='store', required=True)
   parser.add_argument('--root', dest='root', action='store', required=True)
   parser.add_argument('--json', dest='json', action='store', required=True)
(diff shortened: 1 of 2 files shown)


build/fuchsia/sdk.gni
```diff
@@ -243,35 +246,28 @@ template("fuchsia_sdk") {
 
     part_meta_json = read_file(part_meta_rebased, "json")
 
-    # version_history.json does not have a name and thus fails when we try to read it.
-    if (part_meta != "version_history.json") {
+    if (invoker.enabled_parts + [part.type] - [part.type] != invoker.enabled_parts) {
```
akbiggs, line 249: Add a comment explaining this magic.

build/fuchsia/fidl_gen_cpp.py
```diff
@@ -40,7 +40,7 @@ def main():
   parser.add_argument('--sdk-base', dest='sdk_base', action='store', required=True)
   parser.add_argument('--root', dest='root', action='store', required=True)
   parser.add_argument('--json', dest='json', action='store', required=True)
-  parser.add_argument('--fidlgen-root', dest='fidlgen_root', action='store', required=True)
+  parser.add_argument('--fidlgen-output-root', dest='fidlgen_output_root', action='store', required=False)
```
akbiggs, line 43: No action required, but it seems weird that fidlgen_bin is always required even though it won't be used when fidlgen_output_root is not set.

akbiggs (approved):
Thanks for the fix.

build/fuchsia/sdk.gni
chandarrengoog: done.

build/fuchsia/fidl_gen_cpp.py
chandarrengoog: noted. I've changed the code to consider this.

chaselatta (approved):

PR #545, opened 2022-02-03
Build libtess2 with less strict warning flags to fix build

zanderso:
Without this, the `impeller_unittests` target fails to build, giving the error:
```
../../third_party/libtess2/Source/sweep.c:1138:6: error: unused variable 'fixedEdges' [-Werror,-Wunused-variable]
        int fixedEdges = 0;
            ^
1 error generated.
```

### DIFF
--- a/build/secondary/third_party/libtess2/BUILD.gn
+++ b/build/secondary/third_party/libtess2/BUILD.gn
@@ -7,6 +7,9 @@ source_set("libtess2") {
 
   include_dirs = [ "//third_party/libtess2/Include/" ]
 
+  configs -= [ "//build/config/compiler:chromium_code" ]
+  configs += [ "//build/config/compiler:no_chromium_code" ]
+
   sources = [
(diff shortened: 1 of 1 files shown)


chinmaygarde (approved):
Didn't realize I had this uncommitted. Yes, without this patch, the release builds fail because the variable is only used in an assertion. 

Another away to fix this would be to disable the warning in cflags_c. But this way is cleaner.

zanderso:
> Yes, without this patch, the release builds fail because the variable is only used in an assertion.

Huh. There might be a problem with how asserts are enabled, then because I hit the compilation error in a debug build. I will land this and spend a couple minutes looking into it.

chinmaygarde:
That’s odd. Did you mean debug runtime mode or unopt? C asserts are defined
away in unopt which was making the var go unused.

zanderso:
Oh, yeah. I wasn't using --unopt. That was probably it.

PR #573, opened 2022-05-12
Use ld64 for ios_debug

zanderso:
`lld` does not support `-fembed-bitcode-marker`. See https://github.com/llvm/llvm-project/issues/55415.

### DIFF
--- a/build/config/compiler/BUILD.gn
+++ b/build/config/compiler/BUILD.gn
@@ -148,6 +148,13 @@ config("compiler") {
 
     if (enable_bitcode) {
       if (bitcode_marker) {
+        # lld, which is the default, does not support the -bitcode_process_mode
+        # flag that gets plumbed to it as a result of -fembed-bitcode-marker, so
+        # we fall back on ld64.
+        # See: https://github.com/llvm/llvm-project/issues/55415
+        if (!use_xcode) {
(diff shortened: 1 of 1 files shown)


build/config/compiler/BUILD.gn
```diff
@@ -148,6 +148,11 @@ config("compiler") {
 
     if (enable_bitcode) {
       if (bitcode_marker) {
+        # lld, which is the default, does not support the -bitcode_process_mode
+        # flag that gets plumbed to it as a result of -fembed-bitcode-marker, so
+        # we fall back on ld64.
+        # See: https://github.com/llvm/llvm-project/issues/55415
+        ldflags += [ "-fuse-ld=ld" ]
```
dnfield, line 155: nit: maybe only if not xcode? I'm not sure if that's important or not, we don't build this in the buildroot using Xcode but I suppose someone mgiht?

dnfield (approved):
@jmagman fyi

LGTM

build/config/compiler/BUILD.gn
zanderso: I think you'd have to be trying to be trying to compile for ios_sim_arm64 with bitcode to hit this with xcode, but I added a guard anyway.

PR #577, opened 2022-05-28
Buildroot fixes for the Windows arm64 build

zanderso:
Excludes an unsupported linker flag, and fixes a typo.


cbracken (approved):

PR #585, opened 2022-06-24
[gn] add forward variables to compiled_action

XilaiZhang:
context: https://github.com/flutter/engine/pull/34216#issuecomment-1165154820

This commit forwards the metadata variable in `generate_snapshot_bin template`, which is used to generate the `gen_snapshot` engine artifact. Together with the PR on flutter/engine side, this PR eventually aims to embed entitlement metadata in the leaf build targets.

### DIFF
--- a/build/compiled_action.gni
+++ b/build/compiled_action.gni
@@ -81,18 +81,15 @@ template("compiled_action") {
          "compiled_action doesn't take a sources arg. Use inputs instead.")
 
   action(target_name) {
-    if (defined(invoker.visibility)) {
-      visibility = invoker.visibility
-    }
-    if (defined(invoker.testonly)) {
-      testonly = invoker.testonly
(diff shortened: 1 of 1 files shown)


build/compiled_action.gni
```diff
@@ -91,6 +91,8 @@ template("compiled_action") {
       pool = invoker.pool
     }
 
+    forward_variables_from(invoker, ["metadata"])
```
zanderso, line 94: You can use `forward_variables_from` for the three above too as a cleanup:
```suggestion
    forward_variables_from(invoker, ["metadata", "pool", "testonly", "visibility"])
```

XilaiZhang: Very true! newest commit forwards all suitable variables.

zanderso (approved):

PR #599, opened 2022-08-08
GN toolchain setup for host-arm64 Mac and Linux.

rmacnak-google:
This toolchain is enabled only for non-Goma builds.

Copy missing Linux toolchain setup from Dart.

Bug: https://github.com/flutter/flutter/issues/103386

### DIFF
--- a/build/toolchain/linux/BUILD.gn
+++ b/build/toolchain/linux/BUILD.gn
@@ -4,9 +4,13 @@
 
 import("//build/config/sysroot.gni")
 import("//build/toolchain/ccache.gni")
-import("//build/toolchain/clang.gni")
 import("//build/toolchain/gcc_toolchain.gni")
 import("//build/toolchain/goma.gni")
+
+declare_args() {
+  toolchain_prefix = ""
+  # TODO(zra): Add an argument for overriding the host toolchain.
+}
 
 if (use_goma) {
   assert(!use_ccache, "Goma and ccache can't be used together.")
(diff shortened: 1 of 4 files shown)


rmacnak-google:
By itself, this has no effect because flutter/tools/gn won't set host_cpu to arm64.

build/toolchain/linux/BUILD.gn
```diff
@@ -17,29 +21,90 @@ if (use_goma) {
   compiler_prefix = ""
 }
 
+if (host_cpu == "arm64") {
+  rebased_clang_dir =
+      rebase_path("//buildtools/linux-arm64/clang/bin", root_build_dir)
+} else {
```
cbracken, line 27: Is the issue that `host_cpu` isn't reliably set to x64 on x64 machines?

IIRC that may have been the case since I think I tried to do so and it caused an issue that I then didn't dig into to resolve :)

build/toolchain/linux/BUILD.gn
```diff
@@ -89,39 +149,90 @@ gcc_toolchain("clang_x64") {
   is_clang = true
 }
 
-gcc_toolchain("clang_arm64") {
-  if (use_clang_type_profiler) {
-    prefix = rebase_path("//third_party/llvm-allocated-type/Linux_x64/bin",
-                         root_build_dir)
-  } else {
-    prefix = rebase_path("//buildtools/linux-x64/clang/bin", root_build_dir)
+gcc_toolchain("x64") {
```
cbracken, line 152: This fills in some gaps, though we should _probably_ actually test this if we're going to add it as a supported toolchain.

/cc @zanderso for thoughts.

build/toolchain/linux/BUILD.gn
```diff
@@ -89,39 +149,90 @@ gcc_toolchain("clang_x64") {
...
+  ar = "${prefix}ar"
+  ld = cxx
+  strip = "${prefix}strip"
+
+  toolchain_cpu = "x64"
+  toolchain_os = "linux"
+  is_clang = false
+}
+
+gcc_toolchain("riscv32") {
```
cbracken, line 168: Are these primarily for parity with Dart? Similar question to above re: testing since we don't currently build for these today in Flutter.

/cc @zanderso

cbracken (approved):
lgtm for the refactor, curious about Zach's thoughts on testing to avoid potential bitrot.

build/toolchain/linux/BUILD.gn
rmacnak-google: Yes, Flutter's gcc builds are non-function before and after this CL.

build/toolchain/linux/BUILD.gn
rmacnak-google: Mostly that it was the default before so I'm leaving it as the default after. It possible to setup something like Rosetta on Linux using QEMU.

build/toolchain/linux/BUILD.gn
```diff
@@ -4,10 +4,14 @@
 
 import("//build/config/sysroot.gni")
 import("//build/toolchain/ccache.gni")
-import("//build/toolchain/clang.gni")
 import("//build/toolchain/gcc_toolchain.gni")
 import("//build/toolchain/goma.gni")
 
+declare_args() {
+  toolchain_prefix = ""
+  # TODO(zra): Add an argument for overriding the host toolchain.
```
zanderso, line 12: nit: TODOs should link to a GitHub issue.

zanderso (commented):
I don't think we currently have any plans to add Linux arm64 devices to Flutter CI. We currently cross-build the Linux arm64 artifacts that we need. So I would guess that this will bit-rot.

PR #600, opened 2022-08-10
Remove -Wno-unused-but-set-variable.

rmacnak-google:
Whatever toolchain is running on the "Mac iOS Engine" bot doesn't understand this.

Cf. 76a47ff948a26da06b16cd3edb84d30a51d51a5e.


cbracken (approved):
Looks like it might have been removed in a recent clang version?

e.g. https://github.com/bazelbuild/bazel/issues/3622

PR #604, opened 2022-08-24
Do not use Xcode standard library headers for iOS builds

jason-simmons:
See https://github.com/flutter/flutter/issues/110140


build/config/BUILDCONFIG.gn
```diff
@@ -312,7 +312,7 @@ if (!is_clang && using_sanitizer) {
   is_clang = true
 }
 
-use_flutter_cxx = is_clang && (is_linux || is_android || is_mac)
+use_flutter_cxx = is_clang && (is_linux || is_android || is_mac || is_ios)
```
jmagman, line 315: So this works with Xcode 13 clang on Mac iOS Engine Release?

jason-simmons: Confirmed that this works for an `ios_release` build
(`flutter/tools/gn --ios --runtime-mode=release --bitcode --no-goma --no-lto  `)

chinmaygarde (approved):

PR #606, opened 2022-08-26
Enable symbol visibility annotations in libcxxabi on iOS

jason-simmons:
This is required to fix bitcode LTO builds when building the engine
for iOS using libcxx instead of the Xcode standard library headers.

See https://github.com/flutter/flutter/issues/110140


chinmaygarde (approved):

jmagman (approved):
Thanks for the investigation, @jason-simmons!

PR #611, opened 2022-09-06
Override the WASM toolchain to use CanvasKit-specific args

hterkelsen:
Causes the WASM toolchain to use CanvasKit specific arguments.

This allows us to build CanvasKit cross-compiled as part of the normal "host_debug" build, without affecting the normal
Skia build used for Flutter on mobile devices and desktop.

This enables https://github.com/flutter/engine/pull/32510

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/build/toolchain/gcc_toolchain.gni
+++ b/build/toolchain/gcc_toolchain.gni
@@ -280,16 +280,20 @@ template("gcc_toolchain") {
 
     # When invoking this toolchain not as the default one, these args will be
     # passed to the build. They are ignored when this is the default toolchain.
-    toolchain_args = {
-      current_cpu = invoker.toolchain_cpu
-      current_os = invoker.toolchain_os
+    if (defined(invoker.toolchain_args)) {
+      toolchain_args = invoker.toolchain_args
+    } else {
+      toolchain_args = {
+        current_cpu = invoker.toolchain_cpu
+        current_os = invoker.toolchain_os
 
-      # These values need to be passed through unchanged.
-      target_os = target_os
-      target_cpu = target_cpu
+        # These values need to be passed through unchanged.
+        target_os = target_os
+        target_cpu = target_cpu
 
-      if (defined(invoker.is_clang)) {
-        is_clang = invoker.is_clang
+        if (defined(invoker.is_clang)) {
+          is_clang = invoker.is_clang
+        }
       }
     }
 
--- a/build/toolchain/gcc_toolchain.gni
+++ b/build/toolchain/gcc_toolchain.gni
@@ -297,7 +297,6 @@ template("gcc_toolchain") {
       }
     }
 
-
     if (defined(invoker.deps)) {
       deps = invoker.deps
     }
--- a/build/toolchain/gcc_toolchain.gni
+++ b/build/toolchain/gcc_toolchain.gni
@@ -280,22 +280,23 @@ template("gcc_toolchain") {
 
     # When invoking this toolchain not as the default one, these args will be
     # passed to the build. They are ignored when this is the default toolchain.
-    if (defined(invoker.toolchain_args)) {
-      toolchain_args = invoker.toolchain_args
-    } else {
-      toolchain_args = {
-        current_cpu = invoker.toolchain_cpu
-        current_os = invoker.toolchain_os
+    toolchain_args = {
+      current_cpu = invoker.toolchain_cpu
+      current_os = invoker.toolchain_os
 
-        # These values need to be passed through unchanged.
-        target_os = target_os
-        target_cpu = target_cpu
+      # These values need to be passed through unchanged.
+      target_os = target_os
+      target_cpu = target_cpu
 
-        if (defined(invoker.is_clang)) {
-          is_clang = invoker.is_clang
-        }
+      if (defined(invoker.is_clang)) {
+        is_clang = invoker.is_clang
+      }
+
+      if (defined(invoker.extra_toolchain_args)) {
+        forward_variables_from(invoker.extra_toolchain_args, "*")
       }
     }
+
 
     if (defined(invoker.deps)) {
       deps = invoker.deps
--- a/build/toolchain/wasm/BUILD.gn
+++ b/build/toolchain/wasm/BUILD.gn
@@ -26,17 +26,14 @@ gcc_toolchain("wasm") {
   readelf = "readelf"
   nm = "nm"
 
+  is_clang = true
+
   toolchain_cpu = "wasm"
   toolchain_os = "wasm"
 
   link_outputs = [ "{{root_out_dir}}/{{target_output_name}}.wasm" ]
 
-  toolchain_args = {
-    is_clang = true
-    target_os = target_os
-    target_cpu = target_cpu
-    current_cpu = toolchain_cpu
-    current_os = toolchain_os
+  extra_toolchain_args = {
     skia_use_angle = false
     skia_use_dng_sdk = false
     skia_use_dawn = false
--- a/build/toolchain/wasm/BUILD.gn
+++ b/build/toolchain/wasm/BUILD.gn
@@ -26,10 +26,10 @@ gcc_toolchain("wasm") {
   readelf = "readelf"
   nm = "nm"
 
-  is_clang = true
-
   toolchain_cpu = "wasm"
   toolchain_os = "wasm"
+
+  is_clang = true
 
   link_outputs = [ "{{root_out_dir}}/{{target_output_name}}.wasm" ]
 
--- a/build/toolchain/wasm/BUILD.gn
+++ b/build/toolchain/wasm/BUILD.gn
@@ -37,6 +37,7 @@ gcc_toolchain("wasm_for_canvaskit") {
   link_outputs = [ "{{root_out_dir}}/{{target_output_name}}.wasm" ]
 
   extra_toolchain_args = {
+    skia_use_vulkan = false
     skia_use_webgpu = false
     skia_use_libheif = false
     skia_use_libjpeg_turbo_decode = true
@@ -48,6 +49,7 @@ gcc_toolchain("wasm_for_canvaskit") {
     skia_use_lua = false
     skia_use_wuffs = true
     skia_use_zlib = true
+    skia_gl_standard = "webgl"
     skia_enable_gpu = true
     skia_enable_sksl_tracing = false
     skia_use_icu = true
--- a/build/toolchain/wasm/BUILD.gn
+++ b/build/toolchain/wasm/BUILD.gn
@@ -17,7 +17,10 @@ if (use_goma) {
   compiler_prefix = ""
 }
 
-gcc_toolchain("wasm") {
+# This toolchain is specifically for building CanvasKit.
+#
+# It overrides many Skia args in order to facilitate building CanvasKit.
+gcc_toolchain("wasm_for_canvaskit") {
   # emsdk_dir and em_config are defined in wasm.gni.
   ar = "$compiler_prefix$emsdk_dir/upstream/emscripten/emar --em-config $em_config_path"
   cc = "$compiler_prefix$emsdk_dir/upstream/emscripten/emcc --em-config $em_config_path"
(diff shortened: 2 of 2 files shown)


build/toolchain/gcc_toolchain.gni
```diff
@@ -280,16 +280,20 @@ template("gcc_toolchain") {
 
     # When invoking this toolchain not as the default one, these args will be
     # passed to the build. They are ignored when this is the default toolchain.
-    toolchain_args = {
-      current_cpu = invoker.toolchain_cpu
-      current_os = invoker.toolchain_os
+    if (defined(invoker.toolchain_args)) {
+      toolchain_args = invoker.toolchain_args
```
dnfield, line 284: nit: I think instead of letting the invoker specify the toolchain args, can we just let the invoker specify some other object like `extra_toolchain_args` and append them here?

I think GN will let you do that with `+=`, but maybe I'm wrong. It seems strange to let the invoker change the target_os/target_cpu/is_clang this way.

build/toolchain/wasm/BUILD.gn
```diff
@@ -29,7 +29,62 @@ gcc_toolchain("wasm") {
...
+    skia_canvaskit_enable_matrix_helper = false
+    skia_canvaskit_enable_canvas_bindings = false
+    skia_canvaskit_enable_font = true
+    skia_canvaskit_enable_embedded_font = true
+    skia_canvaskit_enable_alias_font = true
+    skia_canvaskit_legacy_draw_vertices_blend_mode = false
+    skia_canvaskit_enable_debugger = false
+    skia_canvaskit_enable_paragraph = true
+    skia_canvaskit_enable_webgl = true
+    skia_canvaskit_enable_webgpu = false
```
dnfield, line 88: We've typically handled these arguments as things specified in `args.gn` set up by the `gn` utility script in the engine repo. Why do we want to move them here? This seems to further complicate buildroot when we're trying to get rid of it/merge it with the engine repo.

hterkelsen: The reason I'm doing it this way is because these skia arguments interfere with the Skia build used by Flutter's mobile engines. An example is enabling `skia_enable_webgl` causes Skia to build with the webgl backend, which is wrong for Flutter mobile.

dnfield: Sure, but does that just mean we should build canvaskit in its own `out` directory with its own args?

I think it'll be confusing to see these values set differently in `args.gn` but actually get overridden here.

hterkelsen: We want to build CanvasKit as part of the `flutter_web_sdk` folder as part of the standard Flutter engine build (e.g. under `host_debug_unopt`)

dnfield: > We want to build CanvasKit as part of the `flutter_web_sdk` folder as part of the standard Flutter engine build (e.g. under `host_debug_unopt`)

@chinmaygarde @zanderso WDYT?

I'm concerned it will be confusing to start introducing artifacts with very different targets/archiectures/build arguments under the same out directory. I'd expect, for example, that we build canvaskit as something like `out/wasm_debug` or something. There are a number of advantages to keeping wasm separate from the `host*` builds too, since it doesn't really differ based on host or target architecture type (so canvaskit.wasm and .js are the same whether you're building on or for x64 or arm64, etc.).

dnfield (changes requested):
Requesting changes at least to give time to get feedback from more of engine team.

build/toolchain/gcc_toolchain.gni
hterkelsen: Just tried this. Unfortunately you can't `+=` a block. But I think I was able to find a way using `forward_variables_from`

build/toolchain/wasm/BUILD.gn
hterkelsen: To give some extra context on why I'm building CanvasKit as part of the `host` build rather than in its own output folder (e.g. `wasm_debug`):

I used to do it that way (building in its own output folder) but it was a bit ugly because (1) I had to have a lot of if-else statements in `tools/gn` which would disable setting arguments in the case of running a WASM build. This was due to those arguments not being used for building CanvasKit. (2) I had to modify the default top-level target to build the Flutter SDK in the normal case, and only build CanvasKit in the WASM case. This is because if certain GN files are even reachable by import when building for WASM (e.g. the Dart SDK BUILD.gn file) it will cause the build to fail (e.g. because Dart doesn't recognize "wasm" as a valid target cpu).

Having a separate output folder for the CanvasKit build was also causing me to write substantial changes to LUCI and flutter_tool since they would have to build a separate output and pass it along for tests and for the Flutter Web SDK.

zanderso: If we think of web as being a different platform, like iOS and Android are different platforms, then giving it its own `out/web_debug` directory makes sense. I don't know the full catalogue of complications that would introduce, though. It would require a careful dance between the GN build and the recipes, at least. Might be worth doing anyway.

Setting that aside, the way we have handled similar situations in the past is by introducing new targets that hardcode the desired configurations. Like here https://github.com/dart-lang/sdk/blob/main/runtime/configs.gni. Then, you depend on the target with hardcoded config with the toolchain specified (`//foo/bar_with_my_config($my_toolchain)`). That way, target/dependency specific configuration doesn't leak into the toolchain definition, which does seem like a layering violation.

There are way more args here than what exists in the Dart repo, though, and I don't know if we can introduce new targets with these configs specified into the Skia build, so the approach that we've followed in the past might not scale to this situation. Is setting all of these args really necessary? Are some of these being set to the default values?

OTOH, if we can't do the above either because there are too many args, or because we can't introduce new targets, we still might improve on the situation by doing something simple like giving the toolchain a more descriptive name, like `"wasm_for_canvaskit"`, and leaving comments like, "Only use this toolchain to build canvaskit". It might also be possible to write some GN `assert`s to enforce that.

hterkelsen: I changed it to be called "wasm_for_canvaskit" and added a comment. I'm still not super familiar with configs but I'm unsure if they allow you to override gn args. I'm overriding them in the toolchain because they differ between the Skia build for Flutter and the Skia build for CanvasKit. I am looking into if I can narrow down the list of args overrides.

yjbanov: I think global toolchain configs were a big bug in our system, and this PR is a step in the right direction. A single invocation of `gn` + `ninja` should be able to build multiple things targeting different architectures and configurations. Otherwise, what you end up is multiple top-level targets that depend on each other, e.g. a host build of the Dart SDK depending on the WASM build of CanvasKit. So now you need yet another build system on top of this to build all the right targets and stitch them together. On top of that there's the issue that @hterkelsen outlines above, where global configs start pushing each other, with web-specific config leaking into mobile versions. I think we should aim for `gn` to be our only build system that's capable of expressing build graphs spanning target architectures.

It also looks like it's appropriate that we put WASM-specific parameters here, because as of right now this is where we configure architecture-specific toolchains. However, I'm a big fan of avoiding maintaining code across repos, so if there's a graceful way to specify these args inside the engine repo, that may be beneficial. If that refactoring requires a lot of time though, I'd prefer we start with this, unblock others waiting on this change (e.g. Skwasm needs this), and then undertake the refactor.

Would that work?

dnfield: It's not clear to me how this will co-exist properly in a single out directory - for example, what will happen when multiple targets want to build the same Skia TUs with different build args?

zanderso: > I think global toolchain configs were a big bug in our system, and this PR is a step in the right direction. A single invocation of `gn` + `ninja` should be able to build multiple things targeting different architectures and configurations. 

This is possible in GN, but I think it would at least require explicitly plumbing a toolchain in many places where it's currently implicit. I wouldn't be surprised if the additional plumbing needed to cross repo boundaries.

> It's not clear to me how this will co-exist properly in a single out directory - for example, what will happen when multiple targets want to build the same Skia TUs with different build args?

Not sure I'm understanding the concern, but specifying a different toolchain makes a different target. Build results for the default toolchain go directly under `out/`. Build results for a different toolchain go under a subdirectory of `out/`, like `out/clang_x64`.

dnfield: Ahh, right, ok

dnfield: @hterkelsen - do you think you could try what Zach is suggesting?

chinmaygarde (changes requested):
Toolchains should not have or enforce opinions about the how they build specific targets. That's the responsibility of `.gn*` files in the targets and their users. I believe toolchain args are also interpreted after `gn args`. So there is no way (please verify) to configure these using the args.gn mechanism and folks will be confused about why their settings are not taking hold when using just this one wasm toolchain. We got into similar messes in the past having global configurations applied to toolchains and those bit-rotting after we updated dependencies. For instance, I think we still have some Valgrind specific global defines left over from the Chromium days even though we have never used it ourselves. This way of specifying toolchain args also does not compose super well.

> I think global toolchain configs were a big bug in our system...

@yjbanov: Toolchain configs are not global. You can specify the toolchain for a target explicitly `//target/to/build:target(//toolchain/to/use:toolchain)`. If you skip the toolchain, the one set by `set_default_toolchain` takes hold.

But what are global are the flags we generally override via declare args. These have been the cause of the whole sordid affair with having multiple out directories.

This patch is perhaps doing too many things here in unblocking the skwasm work and attempting to realize @yjbanovs wish (and mine for whats it's worth) of having a single GN + Ninja step building for multiple architectures. The latter is a much larger effort that would need reworking the `declare args` blocks to use GN template that callers can use. The former can be unblocked right now with a different out directory. We should split the two efforts if necessary.

build/toolchain/wasm/BUILD.gn
```diff
@@ -32,4 +35,56 @@ gcc_toolchain("wasm") {
   is_clang = true
 
   link_outputs = [ "{{root_out_dir}}/{{target_output_name}}.wasm" ]
+
+  extra_toolchain_args = {
+    skia_use_angle = false
```
zanderso, line 40: These can be hooked up to GN args from `args.gn` by introducing new arguments with `declare_args` like `skia_use_angle_for_wasm` (or a better name).

yjbanov: Do you mean that we'd have a file called `args.gn` over in flutter/engine that enumerates all the properties below, and it would override the same properties listed in https://github.com/google/skia/blob/main/modules/canvaskit/canvaskit.gni?

zanderso: No, sorry. I need to think about this PR some more. I'm going to spend some time with it this morning, and see if I can make some better suggestions.

yjbanov: It seems all these attributes are destructured into some if/else block in the skia_wasm_lib used to build canvaskit: https://github.com/google/skia/blob/main/modules/canvaskit/BUILD.gn#L96

Can we convert the Skia rule into a template that takes these same attributes as `invoker.ATTR`?

dnfield: Yes. I think that would be a better way to go here.

zanderso (commented):
I did some local experimentation. When a target is built with the `wasm` toolchain, the default values of GN arguments will be re-evaluated with `is_wasm` set to true. Then, those defaults will be overridden by the contents of `args.gn` (in the Engine build, these are the values setup by the `tools/gn` script.) Finally, `toolchain_args` for the active toolchain applies the final layer of overrides.

This means that the contents of `toochain_args` for the `wasm` toolchain only needs to contain overrides where it wants something different than both the default and what is in `args.gn`.

However, I think before adding entries to the `wasm` toolchain's `toolchain_args`, we should investigate why they're necessary. One example I found was `skia_use_metal`. The default is false, which works for wasm, but the macOS host build overrides it to true. One way to fix this for `wasm` would be to add `skia_use_metal = false` to the `toolchain_args`, but a better way to fix it would be to change reads of `skia_use_metal` to instead be `skia_use_metal && !is_wasm`.

Continuing in that direction, I think we can minimize the leakage.

hterkelsen:
Closing this. After discussion with the engine team we have decided on a different approach: https://github.com/flutter/buildroot/pull/622

PR #618, opened 2022-09-09
Decode gn_run_binary child process output as UTF-8

jason-simmons:
Also accept output with invalid encodings

### DIFF
--- a/build/gn_run_binary.py
+++ b/build/gn_run_binary.py
@@ -20,7 +20,7 @@ path = './' + sys.argv[1]
 args = [path] + sys.argv[2:]
 
 try:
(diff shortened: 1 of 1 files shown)


build/gn_run_binary.py
```diff
@@ -20,7 +20,7 @@
 args = [path] + sys.argv[2:]
 
 try:
-  subprocess.check_output(args, stderr=subprocess.STDOUT, universal_newlines=True)
```
zanderso, line 23: `universal_newlines` has a bunch of side-effects IIRC. Would you mind double-checking the docs to make sure we don't need them?

zanderso (approved):

PR #622, opened 2022-09-13
Override the top level default target in the case of WASM

hterkelsen:
If building for WASM, change the default target to build the WASM artifacts needed by the Web SDK.

This change is necessary for `gn gen` to work when `target_cpu = "wasm"` since the imports from `//flutter/BUILD.gn` lead to a check in `//flutter/shell/platform/BUILD.gn` which throws if the target CPU is not recognized.

We do not (currently?) want to compile the Flutter SDK to WASM. Our only current use case is building CanvasKit for the Flutter Web SDK.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -6,7 +6,7 @@ group("default") {
   testonly = true
   if (target_os == "wasm") {
     deps = [
-      "//flutter/wasm:wasm_artifacts"
+      "//flutter/wasm"
     ]
   } else {
     deps = [
@@ -20,7 +20,7 @@ group("dist") {
 
   if (target_os == "wasm") {
     deps = [
-      "//flutter/wasm:wasm_artifacts"
+      "//flutter/wasm"
     ]
   } else {
     deps = [
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -6,7 +6,7 @@ group("default") {
   testonly = true
   if (target_os == "wasm") {
     deps = [
-      "//flutter/web_sdk/wasm:wasm_artifacts"
+      "//flutter/wasm:wasm_artifacts"
     ]
   } else {
     deps = [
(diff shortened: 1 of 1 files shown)


BUILD.gn
```diff
@@ -4,15 +4,27 @@
 
 group("default") {
   testonly = true
-  deps = [
-    "//flutter",
-  ]
+  if (target_os == "wasm") {
+    deps = [
+      "//flutter/web_sdk/wasm:wasm_artifacts"
```
dnfield, line 9: I think if we change things here, we should just make some specific target for `wasm` e.g. `//flutter/wasm` that is not specific to the web_sdk.

Alternatively, we should do this in `//flutter/BUILD.gn`.

hterkelsen: Putting it in `//flutter/BUILD.gn` causes problems due to targets and imports eventually leading somewhere that will throw due to `target_cpu = "wasm"`. But putting it in `//flutter/wasm/BUILD.gn` sounds good to me. The WASM stuff needs to be quarantined from the normal Flutter SDK build rules because otherwise `gn gen` will error out.

chinmaygarde (approved):

hterkelsen:
FYI here is the corresponding engine change with the `//flutter/wasm/BUILD.gn` rules in it: https://github.com/flutter/engine/pull/32510

BUILD.gn
```diff
@@ -4,15 +4,27 @@
 
 group("default") {
   testonly = true
-  deps = [
-    "//flutter",
-  ]
+  if (target_os == "wasm") {
+    deps = [
+      "//flutter/wasm:wasm_artifacts"
```
dnfield, line 9: Uber not: just make the target wasm.

dnfield (approved):

BUILD.gn
hterkelsen: Done

PR #629, opened 2022-09-26
Public fuchsia_fidl_library template

lin-erik:
`fuchsia_fidl_library` is now public. This allows FIDL protocols in the Engine repo to generate C++ bindings.


naudzghebre (approved):
LGTM

PR #689, opened 2023-03-03
Silence Java deprecation warnings

zanderso:
This PR silences deprecation warnings that are currently emitted while building Java sources for the Android embedder. In general, a successful GN/Ninja build should produce no output. The issue filed in the tracker, and the TODO() in the source will have to serve as sufficient reminder that these deprecations will need to be addressed.

Related https://github.com/flutter/flutter/issues/98602.


camsim99 (approved):
sgtm!

PR #691, opened 2023-03-10
Allow enabling LTO for macOS builds

zanderso:
This will be a no-op because our CI recipes currently explicitly disable LTO. See https://flutter.googlesource.com/recipes/+/refs/heads/main/recipes/engine/engine.py#1085.

After this rolls into flutter/engine, I'll do some `led` runs to gauge the impact to CI cycle time.


jonahwilliams (approved):
LGTM

cbracken (commented):
Thanks!

PR #697, opened 2023-03-23
Properly support .worker.js outputs from emscripten.

eyebrowsoffire:
Also, compile with -g0 in release mode. We were getting unintentionally
bucketed into behavior for Windows and Mac here previously.


hterkelsen (approved):
LGTM

PR #699, opened 2023-03-26
Optimize Wasm for speed instead of size

jason-simmons:
This matches Skia's compile flags for building CanvasKit releases.

See https://github.com/flutter/flutter/issues/123484

### DIFF
--- a/build/config/compiler/BUILD.gn
+++ b/build/config/compiler/BUILD.gn
@@ -874,7 +874,7 @@ config("optimize") {
   } else if (is_android || is_fuchsia) {
     cflags = [ "-Oz" ] + common_optimize_on_cflags  # Favor size over speed.
   } else if (is_wasm) {
-    cflags = [ "-Oz" ]
+    cflags = [ "-O3" ]
   } else {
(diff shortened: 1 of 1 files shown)


yjbanov (approved):
lgtm

yjbanov:
Thanks!

eyebrowsoffire (changes requested):
Do we actually want to do this? What is the binary size difference? What is the runtime performance difference? Have we done any profiling to see if there are particular spots that are slower at -Oz? On the web, binary size is arguable the most important metric. I don't want to use -O3 just because that's what we've done in the past, we should do a real analysis on this.

kevmoo (changes requested):
Let's wait on this. See internal discussion.

yjbanov:
Since this is an easy change, the plan is to try to make `-Oz` work without regressions (at least enough to call the size trade-off a good choice), and if we fail at that, then switch to `-O3`.

eyebrowsoffire (approved):
Fine with landing this for now until we have a better plan.

PR #704, opened 2023-03-30
Build with `-sINLINING_LIMIT` on wasm.

eyebrowsoffire:
This allows us to use LTO but also keep binary size low.


hterkelsen (approved):
LGTM

PR #709, opened 2023-04-10
[dart] Add ntdll.lib for better exception handling in dart

aam:
API added in https://dart.googlesource.com/sdk/+/386a87a8905c48342e1bb9dcbe0ff693690468a7 for better exception handling on Windows needs this library.

### DIFF
--- a/build/config/BUILD.gn
+++ b/build/config/BUILD.gn
@@ -113,6 +113,7 @@ config("default_libs") {
       "dnsapi.lib",
       "iphlpapi.lib",
       "msimg32.lib",
+      "ntdll.lib",
       "odbc32.lib",
       "odbccp32.lib",
       "ole32.lib",


build/config/BUILD.gn
```diff
@@ -113,6 +113,7 @@ config("default_libs") {
       "dnsapi.lib",
       "iphlpapi.lib",
       "msimg32.lib",
+      "ntdll.lib",
```
zanderso, line 116: Shouldn't this go here https://github.com/dart-lang/sdk/blob/main/runtime/BUILD.gn#L148, and should it be limited to only some runtime modes?

zanderso (changes requested):

zanderso (approved):

build/config/BUILD.gn
aam: https://github.com/flutter/flutter/issues/124532

PR #725, opened 2023-04-28
fuchsia: do not read json unless enabled

maruel:
Otherwise this causes GN to read JSON files that are irrelevant.


zanderso (approved):
rslgtm

PR #726, opened 2023-05-04
Use a higher level of debug symbols for Android and other non-Linux/non-Wasm targets

jason-simmons:
The additional symbols are needed so CI can correctly generate the binary size analysis treemaps.


dnfield (approved):

dnfield:
@hterkelsen fyi

PR #728, opened 2023-05-12
Add a gn flag for forcing us to compile wasm with -Oz

eyebrowsoffire:
While we're still building CanvasKit with `-O3`, we want to build skwasm with `-Oz`, so make that a flag that we can tweak.


hterkelsen (approved):
LGTM

PR #729, opened 2023-05-19
Forward inputs from wasm libraries.

eyebrowsoffire:
Some upcoming changes to Skwasm will require us to specify some additional input files that are not sources.


yjbanov (approved):
lgtm

PR #744, opened 2023-06-16
Add malioc runner script

bdero:
Part of https://github.com/flutter/flutter/issues/128989.

The Mali Offline compiler outputs errors in the output file. So add a special execution script that dumps the output file to stdout when malioc fails.

### DIFF
--- a/build/gn_run_malioc.py
+++ b/build/gn_run_malioc.py
@@ -1,0 +1,54 @@
+# Copyright 2013 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+"""Helper script for GN to run malioc.
+
+This is the same as `gn_run_binary.py`, except an extra parameter is included
+for the malioc output file. When the malioc run fails, errors are placed in the
+json output file. This script attempts to read the output file and dump it to
+stdout upon failure.
+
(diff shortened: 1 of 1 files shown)


jonahwilliams (approved):
Looks like python to me!

build/gn_run_malioc.py
```diff
@@ -0,0 +1,54 @@
...
+json output file. This script attempts to read the output file and dump it to
+stdout upon failure.
+
+Run with:
+  python gn_run_malioc.py <binary_name> <output_path> [args ...]
+"""
+
+import json
+import os
+import sys
```
zanderso, line 18: nit: alphabetize imports

build/gn_run_malioc.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2013 The Flutter Authors. All rights reserved.
```
zanderso, line 1: I'm a bit uncertain about the need for this wrapper script since the script that it wraps is already a wrapper around the underlying malioc binary. Is there some way to merge this functionality into the existing wrapper script in the engine repo?

bdero: Oh shoot, I should have read malioc.gni more closely. I mistakenly thought it was running malioc directly.

PR #745, opened 2023-06-16
Fix pretty printing in malioc execution script

bdero:
Whoopies. This fix makes pretty printing actually work. Without this it'll always fail to parse and fall back to raw output.


jonahwilliams (approved):
RSLGTM

PR #746, opened 2023-06-20
Generate new C++ FIDL bindings

tamird:
This is the simplest possible solution for https://fxbug.dev/126295: we don't even bother to generate separate targets; each FIDL library will invoke both fidlgen_cpp and fidlgen_hlcpp, generating sources for both.

Add the hlcpp-generated fidl_test_base.h header while I'm here.


tengyifei:
As someone on Fuchsia and worked on the FIDL bindings this looks good to me. But I don't have approval rights on GitHub.

chaselatta (approved):

PR #748, opened 2023-06-20
Rearrange assertions

tamird:
A previous change made `fidlgen_bin` a list, but did not correctly update this assertion.


zanderso (commented):
rslgtm

zanderso (approved):

PR #751, opened 2023-08-10
Update `dart_roller_helper.py` to explicitly pass `--local-engine-host` to the `flutter` tool.

matanlurey:
Partial work towards https://github.com/flutter/flutter/issues/132245.

These changes are intended to have no semantic change in behavior. Currently the host configuration is _guessed_ by an internal heuristic, but we're switching it to have an explicit configuration instead. In other words, this is future proofing.

_(Worth noting, I'm not actually sure if it's important to have a correct host configuration in release mode)_


jonahwilliams (approved):
RSLGTM

cbracken (approved):
This looks reasonable to me. @bkonyi fyi.

I presume this is still used and that it wasn't specific to @bkonyi's old roll tooling? i.e. guessing we can't just delete this?

bkonyi:
We can delete this script, it's no longer used now that we're using the Skia autoroller. I'll upload a PR when I get into the office today.

PR #752, opened 2023-08-15
Remove old Dart SDK -> Flutter Engine autoroller scripts

bkonyi:
These are no longer used as rolls are done by Skia autoroller infra.


bkonyi:
FYI @aam, just want to confirm these scripts aren't used by the Skia roller (I doubt they are,  but better to be safe than sorry :)).

cc @matanlurey

aam (approved):

cbracken (commented):
Thanks!

PR #759, opened 2023-09-01
Add a toolchain for Android for building with api level 26

zanderso:
Needed for https://github.com/flutter/engine/pull/45401

PR #761, opened 2023-09-05
Add comment on arm64_apilevel26 Android GN toolchain

zanderso:
Should have had this in https://github.com/flutter/buildroot/pull/759.

### DIFF
--- a/build/toolchain/android/BUILD.gn
+++ b/build/toolchain/android/BUILD.gn
@@ -178,6 +178,10 @@ android_gcc_toolchains_helper("arm64") {
   toolchain_cpu = "aarch64"
 }
 
+# This toolchain should only be used to build the target
+# //third_party/vulkan_validation_layers. This is because vulkan validation
+# layers requires API level >= 26, but Flutter officially supports down to API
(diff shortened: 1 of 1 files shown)


build/toolchain/android/BUILD.gn
```diff
@@ -178,6 +178,10 @@ android_gcc_toolchains_helper("arm64") {
   toolchain_cpu = "aarch64"
 }
 
+# This toolchain should only be used to build the target
+# //third_party/vulkan_validation_layers. This is because vulkan validation
+# layers requires API level >= 26, but Flutter officially supports down to API
+# level 22, which is the default value of the android_api_level argument.
```
gaaclarke, line 184: might want to mention something about how this is safe only because of runtime checks before loading the dynamic library

gaaclarke (approved):

build/toolchain/android/BUILD.gn
zanderso: Yeah good point. Adding a comment to that effect in https://github.com/flutter/engine/pull/45467.

PR #779, opened 2023-10-16
Adds support for building for Android riscv64

zanderso:


goderbauer:
@zanderso says this is currently blocked behind deprecating support for API level 19.

PR #799, opened 2023-12-08
Allow `colorama` to be in 1 of 2 paths temporarily.

matanlurey:
Would allow landing https://github.com/flutter/engine/pull/48842.

I have no idea if this is safe to do, and no real way to test it, so hoping code review to the rescue!


dnfield (approved):
LGTM, consider filing a bug to go with the TODO

matanlurey:
I'm actually going to go with @zanderso's proposal to just inline the variables and delete the dependency. Thanks for the review Dan!

PR #802, opened 2023-12-12
Remove unused `jsoncpp` override.

matanlurey:
I believe this was removed in https://github.com/flutter/engine/pull/8230.

I'll verify this by rolling buildroot and removing the `BUILD.gn` files from the engine.


jason-simmons (commented):
It looks like the ANGLE build scripts depend on this variable.

See https://chromium.googlesource.com/angle/angle.git/+/refs/heads/main/BUILD.gn#1684

matanlurey:
Ah rip thanks!

PR #806, opened 2023-12-16
Strip absolute paths that may be found in depfiles generated by Goma builds

jason-simmons:
This workaround was originally added for RBE.  Builds configured with --goma can also output depfiles that contain the same absolute path prefix seen in RBE builds.


zanderso (approved):

zanderso (commented):
I believe to be portable, the sed command will need to look like:
```
sed -i'' -e 's@/b/f/w@../..@g' $depfile
```

https://unix.stackexchange.com/questions/92895/how-can-i-achieve-portability-with-sed-i-in-place-editing

PR #809, opened 2024-01-05
Make the workaround for absolute paths in depfiles compatible with macOS

jason-simmons:
See https://github.com/flutter/engine/pull/49542

### DIFF
--- a/build/toolchain/gcc_toolchain.gni
+++ b/build/toolchain/gcc_toolchain.gni
@@ -122,6 +122,12 @@ template("gcc_toolchain") {
       coverage_flags = "-fprofile-instr-generate -fcoverage-mapping"
     }
 
+    if (host_os == "mac") {
+      sed_in_place_flag = "-i ''"
+    } else {
+      sed_in_place_flag = "-i"
+    }
+
     tool("cc") {
       depfile = "{{output}}.d"
       command = "$cc -MD -MF $depfile {{defines}} {{include_dirs}} {{cflags}} {{cflags_c}} $coverage_flags -c {{source}} -o {{output}}"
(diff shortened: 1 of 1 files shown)


jmagman (approved):
On master on macOS:
```
/fml/platform/android/fml.scoped_java_ref.o && sed -i 's@/b/f/w@../..@g' obj/flutter/fml/platform/android/fml.scoped_java_ref.o.d
sed: 1: "obj/flutter/fml/platfor ...": invalid command code o
```
On this PR: success.

build/toolchain/gcc_toolchain.gni
```diff
@@ -122,14 +122,20 @@ template("gcc_toolchain") {
       coverage_flags = "-fprofile-instr-generate -fcoverage-mapping"
     }
 
+    if (host_os == "mac") {
```
zanderso, line 125: `-i''` that is, an empty string with no space after the `-i` should work on both, but maybe it's better to be explicit like this.

zanderso (approved):

build/toolchain/gcc_toolchain.gni
jason-simmons: In my local testing on macOS `-i''` did not work.

It looked like it was being parsed as equivalent to `-i`, causing the next argument (the  sed command) to be used as the backup file extension.

zanderso: Ah, maybe it only works with `-e` before the command to delineate the empty string from the command like: `sed -i'' -e 's@/b/f/w@../..@g' $depfile`.

jason-simmons: I tried that originally but found that it was creating a backup file with the extension `-e` added to the filename

zanderso: Hahaha. Okay. I'm glad you checked =)

jmagman:
New errors in roll:
```
ERROR at //build/toolchain/gcc_toolchain.gni:128:27: Assignment had no effect.
      sed_in_place_flag = "-i"
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8759705668090838113/+/u/gn_--web_--runtime-mode_release_--no-goma_--no-lto/stdout

https://github.com/flutter/engine/pull/49542

jason-simmons:
Fix for the unused variable error: https://github.com/flutter/buildroot/pull/810

PR #810, opened 2024-01-06
Move the sed in place edit flag outside the gcc_toolchain template

jason-simmons:
This avoids GN warnings in targets such as Wasm where this variable is never used.


zanderso (approved):

zanderso:
Merging to try to get this roll unblocked.

PR #813, opened 2024-01-11
Add fidl_binding_deps support from the Fuchsia SDK json files

mbrase:
The GN rules that parse the SDK json files previously only supported the `fidl_deps` field for determining which fidl dependencies to depend on. This change adds support for the newer `fidl_binding_deps` field, which contains information about what flavor of bindings are needed (hlcpp vs natural). Since our fidl GN targets always generate both hlcpp and natural bindings, we don't need to check the binding_type in the json file.


tarobins:
lgtm

tarobins (approved):

PR #798, opened 2023-12-06
Add FIDL HLCPP tables.c file to the build

mk12:
This is a continuation of #795. This tables.c file is currently empty, but we will migrate the fidlc coding tables to it. See https://fxrev.dev/953989 for the full soft transition plan.

PR #821, opened 2024-01-26
Remove old FIDL tables.c file from build

mk12:
Now that #798 added the new tables.c file, and https://fxrev.dev/946309 moved the content from the old file to the new file, we can stop generating the old file (i.e. stop passing --tables to fidlc).

See https://fxbug.dev/42115242 for context.


jrwang (approved):

mk12:
I manually tested that this works:

```sh
$ENGINE_DIR/flutter/tools/gn --fuchsia --no-lto --fuchsia-cpu=arm64 --runtime-mode=debug --goma --unoptimized
ninja -C $ENGINE_DIR/out/fuchsia_debug_unopt_arm64 flutter/shell/platform/fuchsia
```

I also confirmed GN errors and linker errors when I changed the diff to be incorrect, so I know I'm building the right thing.

PR #822, opened 2024-01-30
Updates for Android NDK version 26

zanderso:
This PR adjusts the buildroot for Android NDK version 26. This NDK version only supports clang, so it contains some cleanups of old non-clang pathways.

The companion flutter/engine PR is https://github.com/flutter/engine/pull/48254.

### DIFF
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -5,15 +5,6 @@
 # This file contains common system config stuff for the Android build.
 
 if (is_android) {
-  has_chrome_android_internal =
-      exec_script("//build/dir_exists.py",
-                  [ rebase_path("//clank", root_build_dir) ],
-                  "string") == "True"
-
-  if (has_chrome_android_internal) {
-    import("//clank/config.gni")
-  }
-
   if (!defined(default_android_sdk_root)) {
     default_android_sdk_root = "//third_party/android_tools/sdk"
     default_android_sdk_version = "34"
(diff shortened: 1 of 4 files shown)


reidbaker (approved):
RSLGTM.

build/config/android/config.gni
```diff
@@ -92,69 +67,13 @@ if (is_android) {
 
   zipalign_path = "$android_sdk_build_tools/zipalign"
 
-  # Subdirectories inside android_ndk_root that contain the sysroot for the
-  # associated platform.
-  if (current_cpu != "x64" && current_cpu != "arm64") {
-    android_api_level = 19
-  }
```
dnfield, line 99: I think this needs to stay and get changed to 21. The default above is 22.

zanderso: Good catch.

build/config/android/config.gni
```diff
@@ -25,22 +16,6 @@ if (is_android) {
     android_sdk_version = default_android_sdk_version
     android_sdk_build_tools_version = default_android_sdk_build_tools_version
 
-    android_default_keystore_path =
```
dnfield, line 28: nit: if you're removing these, maybe remove them from the declare_args block above too?

zanderso: Could you clarify? These deletions are the deletions from the `declare_args` block.

dnfield (approved):
LGTM

PR #827, opened 2024-02-26
Limit GN header checking in //flutter to the first-party engine subsystems

jason-simmons:
Some components that do not support the GN header checks (such as the Dart SDK) are being moved into //flutter/third_party.  These directories should be excluded from "gn check".

See https://github.com/flutter/flutter/issues/143335

### DIFF
--- a/.gn
+++ b/.gn
@@ -16,11 +16,12 @@ secondary_source = "//flutter/build/secondary/"
 # The set of targets known to pass 'gn check'. When all targets pass, remove
 # this.
 check_targets = [
-  "//dart/*",
-  "//flow/*",
-  "//flutter/*",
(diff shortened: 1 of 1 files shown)


.gn
```diff
@@ -18,7 +18,14 @@ secondary_source = "//flutter/build/secondary/"
 check_targets = [
```
zanderso, line 18: Should we take the opportunity to remove the entries here that don't exist anymore?

zanderso (approved):
lgtm

.gn
jason-simmons: done

chinmaygarde (approved):

PR #844, opened 2024-04-10
Output .js files from wasm modules as ES6 modules.

eyebrowsoffire:
Buildroot changes needed for https://github.com/flutter/engine/pull/52023

### DIFF
--- a/build/config/compiler/BUILD.gn
+++ b/build/config/compiler/BUILD.gn
@@ -296,6 +296,7 @@ config("compiler") {
       # Reduces global namespace pollution.
       "-s",
       "MODULARIZE=1",
+      "-sEXPORT_ES6",
 
       # Always produce a symbol map so that we can map crash traces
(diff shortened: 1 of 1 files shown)


build/config/compiler/BUILD.gn
```diff
@@ -296,6 +296,7 @@ config("compiler") {
       # Reduces global namespace pollution.
       "-s",
       "MODULARIZE=1",
+      "-sEXPORT_ES6",
```
ditman, line 299: Why not

```
"-s",
"EXPORT_ES6",
```

Like the other options? (is it because it doesn't have `=`?)

ditman (approved):
This makes sense!

build/config/compiler/BUILD.gn
eyebrowsoffire: I think I just made an arbitrary choice. I can make it consistent with the others.

PR #854, opened 2024-05-08
Adds RBE local and remote wrappers for Linux and Android

zanderso:
For https://github.com/flutter/flutter/issues/147750. This change also disables remote linking in anticipation of changing to differently sized workers in the worker pool.

Validated in https://github.com/flutter/engine/pull/52675.


gaaclarke (approved):
lgtm

PR #857, opened 2024-05-13
Pass stack protection flag on Android

zanderso:
For b/338202330.

This increases binary size of libflutter.so in android_release_arm64 by ~55KB (+0.5%) uncompressed, and by ~30KB (+0.6%) gzip'd.

Tested in https://github.com/flutter/engine/pull/52774


zanderso:
cc @jiahaog

jason-simmons (approved):

PR #884, opened 2024-08-07
[macOS] Fix COPY step on macOS 15 Sequoia

cbracken:
macOS build fails on Sequoia beta 2 and later (verified in public beta 3 as well) hosts during ninja `COPY` steps when copying files with the `restricted` extended attribute set. As part of our build, we copy the `/System/Library/Fonts/Apple Color Emoji.ttc` which is one such file.

The problem appears to be due to the `-a` (archive) option, which is an alias for `-RpP`. Of these, the `-p` option appears to be the source of the problem. This flag attempts to preserve file metadata including Access Control Lists and Extended Attributes. However, the `restricted` extended attribute is not settable when System Integrity Protection (SIP) is enabled, which it is by default.

This appears to be a change in the implementation of `cp` in macOS 15 Sequoia beta 2 and later. Previous versions would fail to copy the attribute, but continue without failing. The current version appears to fail if it cannot set a given extended attribute.

Two alternatives to the previous `cp -af` step:
* Use `cp -Rf`. This succeeds but file metadata such as creation time, modification time, etc. are not preserved.
* Use `rsync -arl`. rsync's -a flag also copies metadata but on a best-efforts basis and doesn't error out when it fails to set the restricted attribute.

Of the two, the latter more closely approximates the previous behaviour.

Fixes: https://github.com/flutter/flutter/issues/152978

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read the [Flutter Style Guide] _recently_, and have followed its advice.
- [X] I signed the [CLA].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I added new tests to check the change I am making, or this PR is [test-exempt].
- [X] All existing and new tests are passing.

[Contributor Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#overview
[Tree Hygiene]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md
[test-exempt]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/blob/master/docs/contributing/Style-guide-for-Flutter-repo.md
[breaking change policy]: https://github.com/flutter/flutter/blob/master/docs/contributing/Tree-hygiene.md#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/blob/master/docs/contributing/Chat.md

### DIFF
--- a/build/toolchain/mac/BUILD.gn
+++ b/build/toolchain/mac/BUILD.gn
@@ -222,7 +222,7 @@ template("mac_toolchain") {
     }
 
     tool("copy") {
-      command = "ln -f {{source}} {{output}} 2>/dev/null || (rm -rf {{output}} && cp -af {{source}} {{output}})"
+      command = "ln -f {{source}} {{output}} 2>/dev/null || (rsync -a --delete {{source}} {{output}})"
       description = "COPY {{source}} {{output}}"
     }
 


zanderso (approved):

build/toolchain/mac/BUILD.gn
```diff
@@ -222,7 +222,7 @@ template("mac_toolchain") {
     }
 
     tool("copy") {
-      command = "ln -f {{source}} {{output}} 2>/dev/null || (rm -rf {{output}} && cp -af {{source}} {{output}})"
+      command = "ln -f {{source}} {{output}} 2>/dev/null || (rm -rf {{output}} && rsync -arl {{source}} {{output}})"
```
jmagman, line 225: Instead of `(rm -rf {{output}} && rsync -arl {{source}} {{output}})` 
1. `man rsync` says `-a` is the same as `-rlptgoD` so do we need the `rl`?
2. Instead of the `rm -rf` we can use `rsync --delete` so it will only copy what's needed, but delete anything extraneous (this is what we do in the tool)
https://github.com/flutter/flutter/blob/0a7f8af6d14696e9086cdd390792f7eb8a9cefa5/packages/flutter_tools/lib/src/base/os.dart#L458-L461

```suggestion
      command = "ln -f {{source}} {{output}} 2>/dev/null || (rsync -a --delete {{source}} {{output}})"
```

jmagman: Also, do you know what's up with the symlink part?

cbracken: > so do we need the `rl`?

Good catch; nope, safe to remove.

> do you know what's up with the symlink part?

The code is trying to be clever. The logic is basically:
1. If possible, create a hard link (notice it's not `-s`), deleting any existing file.
2. If that's not possible (e.g. the source file is on a different volume), then `rm -rf` the previous file and copy the file.

This logic definitely predates APFS. With APFS, copies are free (well you still pay the cost of the inodes) and the copy is materialised only if you edit the file. i.e. APFS is effectively copy-on-write.

> Instead of the `rm -rf` we can use `rsync --delete`

Looks like the default behaviour is `--delete-before` when `--delete` is specified, so I _think_ this would preserve the existing behaviour exactly. Will do a quick test locally just out of paranoia.

cbracken: Re: APFS, looking closer at the `cp` code, `cp` only takes advantage of cloned inodes if the [`-c` option is passed](https://github.com/apple-oss-distributions/file_cmds/blob/414172a34456f1b54807c9c0aa5d9200a585333f/cp/utils.c#L289-L297), but falls back to `copyfile()` if `clonefile()` fails, which presumably is under the same conditions as hardlinking would -- i.e. source and destination on a different volume.

I'd have to look to see if `rsync` takes advantage of this or not; there's no mention in the manpage.

cbracken: Based on a really quick grep of `rsync` it doesn't appear Apple is using `clonefile` or passing `COPYFILE_CLONE` etc. to `copyfile` flags so tempted to keep the `ln`.

PR #912, opened 2024-10-11
Use xcrun to discover Apple SDK paths

zanderso:
Tested in https://github.com/flutter/engine/pull/55818

### DIFF
--- a/build/config/ios/ios_sdk.py
+++ b/build/config/ios/ios_sdk.py
@@ -93,11 +93,10 @@ def main(argv):
 
   for sdk in sdks:
     command =  [
-      'xcodebuild',
-      '-version',
-      '-sdk',
+      'xcrun',
+      '--sdk',
       sdk,
-      'Path'
+      '--show-sdk-path',
     ]
     sdk_output = run_command_with_retry(command, timeout=300)
     if symlink_path:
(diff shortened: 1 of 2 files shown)


build/config/ios/ios_sdk.py
```diff
@@ -99,7 +104,16 @@ def main(argv):
       sdk,
       'Path'
     ]
-    sdk_output = run_command_with_retry(command, timeout=300)
+    if args.xcode is not None:
+      if sdk in ['iphoneos']:
+        sdk_output = os.path.join(args.xcode, 'iPhoneOS.platform/Developer/SDKs/iPhoneOS17.0.sdk')
```
jmagman, line 109: I really don't want the version hardcoded since this is coming from the Xcode sdk version in the .ci.yaml.

Can we try swapping the ios_sdk and mac_sdk python script to `xcrun --sdk <sdk> --show-sdk-path` instead, and see if it still times out?

zanderso: I agree, and I'll try that, yeah.

zanderso:
@jmagman https://github.com/flutter/engine/pull/55818 is all green, so I think this is working.

jmagman (approved):
Cross referencing with https://github.com/flutter/flutter/issues/94513 re: the first attempt to use hard-coded strings.

LGTM, thank you, @zanderso!

PR #916, opened 2024-10-30
Changes a few compile options for the new emscripten (3.1.70)

eyebrowsoffire:
* Specifying the environments explicitly avoids emitting extra code to handle node.js and other environments.
* Newer versions of emscripten don't actually produce a .worker.js file anymore, even in pthread builds.


harryterkelsen (approved):
LGTM!

PR #918, opened 2024-10-31
Reland "Changes a few compile options for the new emscripten (3.1.70)" (#917)

eyebrowsoffire:
This relands the emscripten config changes that had to be reverted previously. This is to support the new version of emscripten introduced in this engine PR: https://github.com/flutter/engine/pull/56282


harryterkelsen (approved):
LGTM

